PR #29, opened 2019-09-02
add zt-cv-dmrg

jiangtong1000:
try first pull request, add correction vector DMRG for zero temperature spectra.
to do :

[ ] finite temperature spectra as well as transport


renormalizer/cv/zerot.py
```diff
@@ -0,0 +1,345 @@
...
+        self.init_mps()
+        self.cv_mps = Mps.random(
+            self.mol_list, 1-self.nexciton, self.m_max, percent=1.0)
+
+        total_num = 1
+        spectra = []
+
+        # Here did not implement cpu-parallelization
+        for omega in self.freq_reg:
+            print('omega', omega)
```
liwt31: If this is for debugging, I'd suggest to comment this out. If this is for logging, using the `logging` module would be more consistent with other parts of the project.

renormalizer/cv/zerot.py
```diff
@@ -0,0 +1,345 @@
...
+        mat_a = scipy.sparse.linalg.LinearOperator((nonzeros, nonzeros), matvec=hop)
+        if num in [1, 2]:
+            x, info = scipy.sparse.linalg.cg(mat_a, vec_b.asnumpy(), atol=0)
+        else:
+            x, info = scipy.sparse.linalg.cg(mat_a, vec_b.asnumpy(), tol=1.e-5, x0=guess, M=pre_a_mat, atol=0)
+        try:
+            assert info == 0
+        except:
+            print('not convergd')
+            pass
```
liwt31: Just nagging:
```
if info != 0:
    print("not converged")
```
The current version is, of course, fine. You can leave it here.

jiangtong1000: yes, could be better

renormalizer/cv/zerot.py
```diff
@@ -0,0 +1,351 @@
...
+            if direction == "left":
+                self.cv_mps[isite - 1] = x
+                self.cv_mps[isite - 2] = compx
+            else:
+                self.cv_mps[isite - 2] = x
+                self.cv_mps[isite - 1] = compx
+            self.cv_mps.qn[isite - 1] = xqn
+        return l_value
+
+    def initialize_LR(self, direction):
```
liwt31: I'm not sure about this, but have you looked at the `Environ` class in `renormalizer.mps.lib`? It is built to deal with this "LR" thing. If it does not satisfy your need, is it possible to make improvements to the `Environ` class so that other people can benefit from your code? Thanks.

liwt31:
Thanks for your contribution! Before it's merged I think it'll be better to clean up the commit history as it contains debug info such as `wtli tweak`. I can help you with this once the PR is ready to merge.

renormalizer/cv/zerot.py
```diff
@@ -0,0 +1,351 @@
...
+                                           first_R)
+            cout = ax[qnmat == constrain_qn].reshape(nonzeros, 1).asnumpy()
+            return cout
+
+        mat_a = scipy.sparse.linalg.LinearOperator((nonzeros, nonzeros), matvec=hop)
+        if num in [1, 2]:
+            x, info = scipy.sparse.linalg.cg(mat_a, vec_b.asnumpy(), atol=0)
+        else:
+            x, info = scipy.sparse.linalg.cg(mat_a, vec_b.asnumpy(), tol=1.e-5, x0=guess, M=pre_a_mat, atol=0)
+        logger.info('hop times:{count[0]}')
```
liwt31: `f` missing here

renormalizer/cv/zerot.py
jiangtong1000: Good suggestions! It seems that the zero temperature structure looks much similar as ground state calculation, which will be easy by improving the Environ class a little bit.
However, for finite temperature code, firstly, it is not a <\psi|H\psi>-like structure, which could be more complicated to generalize a common structure for all terms. Any smart ideas?

image

liwt31: This is too complicated for me :stuck_out_tongue_closed_eyes:. I'd say let's leave it here, or we can hear what @jjren say about this. This PR requires his approvement anyway.

jjren: If this kind of matrix contraction pattern shares some features with the other contraction pattern, I think it is better to put them together. However, if the pattern is independent and only for CV approach, it is ok to leave it here.

jjren:
From now on, we'd better add important and necessary comments in the source file. For example, the reference that describes the algorithm, how to use your CV algorithm (API).

liwt31:
@jiangtong1000 I'm ok with your PR but it seems you need to add necessary comments firstly. After you've done that I think the PR is ready to be merged.

jiangtong1000:
sure~

liwt31:
Thanks for the comments. However, an essential part which is *how to use your CV algorithm (API)* is missing. To be specific, It'll be best if you can add comments about the arguments of your class and functions, at least for public interfaces (such as the `__init__` function).

renormalizer/cv/zerot.py
liwt31: Let's leave it here and we can refactor the code if we need a similar feature in the future.

jiangtong1000:
> Thanks for the comments. However, an essential part which is _how to use your CV algorithm (API)_ is missing. To be specific, It'll be best if you can add comments about the arguments of your class and functions, at least for public interfaces (such as the `__init__` function).

Thanks Weitang. Actually, I have done that at the test file, which may be able to illustrate the usage？

jiangtong1000:
add doc, just an initial demo version @jjren @liwt31 , you may want to link it to github pages to make it visible to us

renormalizer/cv/__init__.py
```diff
@@ -0,0 +1,6 @@
+# -*- coding: utf-8 -*-
+# Author: Jiajun Ren <jiajunren0522@gmail.com>
```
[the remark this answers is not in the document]
jiangtong1000: 这个是当时直接从spectra里的test文件夹直接copy过来的😅没仔细看

jiangtong1000:
> > add doc, just an initial demo version @jjren @liwt31 , you may want to link it to github pages to make it visible to us
> 
> I think only the source file should be included, like pyscf. @liwt31

fine, another repo renormalizer.github.io is required,

liwt31:
I think it's working quite well. We can now see the doc [here](https://shuaigroup.github.io/Renormalizer/). The next step should be finding out a way to automatically update the pages I suppose.

Issue #30, opened 2019-09-02
The 3rd-party libraries used in Renormalizer 


jjren:
There are several 3rd-party libraries used in Renormalizer.
I would like to list them all and figure out their roles in the whole package. 
Please help to check and finish it. @liwt31 

1. /lib/davidson.py       
https://github.com/pyscf/pyscf/blob/master/pyscf/lib/linalg_helper.py
davidson diagonalization 

2. lib/_ivp
scipy.integrate
solve initial value problem 
(I guess this file is modified for Cupy, right? If the scipy.integrate is upgraded, what should we do?)

3. lib/krylov.py
https://github.com/cmendl/pytenet/blob/master/pytenet/krylov.py
Krylov space method to calculate e^{iH\tau} \Psi

4. utils/qutip_utils.py
It seems that qutip_utils.py calculates the matrix element of some basic operators.

liwt31:
As for 2: it's heavily modified to be compatible with CuPy. In the near future when [NEP18](https://numpy.org/neps/nep-0018-array-function-protocol.html) is enabled, we might be able to use NumPy array and CuPy array interchangeably and the modified lib in `lib/_ivp` will become unnecessary.
As for 4: it's not a third party lib. I wrote them actually.

jjren:
I mean some qutip function like qutip.destroy, qutip.identity.

liwt31:
> I mean some qutip function like qutip.destroy, qutip.identity.

This is okay, otherwise, we need to include NumPy and lots of other packages either.

Issue #28, opened 2019-09-02
Proper license should be added


jjren:
I think a proper license should be added. What do you think about it? @liwt31

liwt31:
Some reasonable choices:
- MIT. Used in [pytenet](https://github.com/cmendl/pytenet/blob/master/LICENSE)(BSD-2 seems to be similar with MIT), [evoMPS](https://github.com/amilsted/evoMPS/blob/master/LICENSE.txt)
- Apache 2.0. Used in [pyscf](https://github.com/pyscf/pyscf/blob/master/LICENSE), [quimb](https://github.com/jcmgray/quimb/blob/develop/LICENSE.txt), [ITensor](https://github.com/ITensor/ITensor/blob/v3/LICENSE), [TensorNetwork](https://github.com/google/TensorNetwork/blob/master/LICENSE)
- GPL. Used in [tenpy](https://github.com/tenpy/tenpy/blob/master/LICENSE)

MIT is the freest license, it almost has no limitations to users. Apache is slightly more strict, requiring users to state their changes. GPL is the most strict one, requiring users to use GPL license.

I'm preferring Apache because it seems to be the most popular one.

liwt31:
Another issue about licenses is that we need to include licenses of the third-party libs that we use, which is written in the terms of their license. I think I can figure this out when I have some time.

jjren:
Apache 2.0 seems appropriate. 
I just opened an issue #30 to list all the 3rd-party libraries in Renormalizer.

PR #33, opened 2019-09-05
add 3rd party lib licenses

liwt31:
Add 3rd party lib licenses according to #28 and #30.
- For davidson from pyscf, apache2 is used. Thus we include a license and state changes.
- For integrator from scipy, BSD3 is used . Thus we include their license.
- For krylov expm from pytenet, BSD2 is used. Thus we include their license.


renormalizer/mps/solver.py
```diff
@@ -7,7 +7,7 @@
 import scipy
 
 
-from renormalizer.lib.davidson import davidson
+from renormalizer.lib import davidson
```
jjren: why is this import different from the style in /lib/__init__.py 
"from renormalizer.lib.davidson.davidson import davidson"?

liwt31: In `lib/init.py` it can be understood as `from .davidson.davidson import davidson`. 
After `lib/init.py` is executed, the `davidson` function appears under `renormalizer.lib` and we can import this way here.

jjren: got it.

PR #34, opened 2019-09-07
add distance test and fix a few bugs

liwt31:
- Add tests for distance calculation just in case. The current implementation seems ok.
- Fix a few bugs. The one concerning `contract_compress_config` is very serious and may cause P&C with fixed M to behave unexpectedly.


liwt31:
@jjren conflicts resolved but I need your approval again.

Issue #25, opened 2019-08-31
Compress the test suite


liwt31:
Currently we have too many tests. A successful build takes 37 mins according to [travis CI](https://travis-ci.com/jjren/Renormalizer/builds/124459803).

It would be ideal if we can cut or modify some of the tests. I think ~20 min build time is more reasonable.

Tests that take more than 10 s according to [travis CI](https://travis-ci.com/jjren/Renormalizer/builds/124459803):
```
110.64s call     renormalizer/spectra/tests/test_spectra.py::test_zero_exact_emi
100.36s call     renormalizer/mps/tests/test_quasiboson.py::test_quasiboson_solver[value0]
88.82s call     renormalizer/spectra/tests/test_hybrid_spectra.py::test_hybrid_emi[1-mol_list1-ZeroExactEmi.npy-0.01]
88.52s call     renormalizer/spectra/tests/test_hybrid_spectra.py::test_hybrid_emi[2-mol_list3-ZeroExactEmi.npy-0.01]
71.81s call     renormalizer/mps/tdh/tests/test_tdh.py::test_TDH_ZT_emi
67.89s call     renormalizer/mps/tests/test_evolve.py::test_finite_t_spectra_emi_TDVP[EvolveMethod.tdvp_mctdh_new-85-4-None-0.01-2]
67.41s call     renormalizer/spectra/tests/test_hybrid_ft_spectra.py::test_ft_hybrid_dmrg_tdh[mol_list0-abs-hybrid_FT_abs_pure.npy]
64.29s call     renormalizer/spectra/tests/test_hybrid_spectra.py::test_hybrid_emi[2-mol_list2-hybrid_ZTemi_prop.npy-0.001]
63.15s call     renormalizer/spectra/tests/test_hybrid_spectra.py::test_Exact_Spectra_hybrid_TDDMRG_TDH
62.44s call     renormalizer/spectra/tests/test_hybrid_spectra.py::test_hybrid_emi[1-mol_list0-hybrid_ZTemi_prop.npy-0.001]
53.94s call     renormalizer/mps/tests/test_mpdm.py::test_thermal_prop[EvolveMethod.tdvp_ps-True-None-mol_list1-0.0853413581416-occ_std1-0.005]
53.36s call     renormalizer/spectra/tests/test_hybrid_ft_spectra.py::test_ft_hybrid_dmrg_tdh[mol_list2-emi-hybrid_FT_emi_pure.npy]
[...]

PR #41, opened 2019-09-29
improve tdvp vmf

liwt31:
- Calculate S according to previous calculated S to improve efficiency.
- Hack into scipy ivp so that during one call of `solve_ivp` only one y is stored.
- Fix tests by larger regularization epsilon

### DIFF
--- a/renormalizer/mps/mps.py
+++ b/renormalizer/mps/mps.py
@@ -696,11 +696,7 @@ class Mps(MatrixProduct):
         else:
             coef = 1j
 
-        if not self.check_left_canonical():
-            self.move_qnidx(0)
-            self.left = True
-            self.canonicalise()
-            assert self.check_left_canonical()
+        self.ensure_left_canon()
         
         # `self` should not be modified during the evolution
         if imag_time:
@@ -721,7 +717,9 @@ class Mps(MatrixProduct):
 
             environ = Environ(mps, mpo, "L")
             environ.write_r_sentinel(mps)
-            
+
+            # the first S
+            S = ones([1, 1], dtype=mps.dtype)
             # calculate hop_y: from right to left
             hop_y = xp.empty_like(y)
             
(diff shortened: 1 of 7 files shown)


renormalizer/mps/tests/test_evolve.py
```diff
@@ -35,6 +35,8 @@ def test_ZeroTcorr_TDVP(method, evolve_dt, nsteps, use_rk, cmf_or_midpoint, rtol
     evolve_config.tdvp_ps_rk4 = use_rk
     evolve_config.tdvp_mctdh_cmf = cmf_or_midpoint
     evolve_config.tdvp_mu_midpoint = cmf_or_midpoint
+    if method is EvolveMethod.tdvp_vmf:
+        evolve_config.reg_epsilon = 1e-5
```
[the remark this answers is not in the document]
liwt31: Sure. But the goal here is to make the test work, no matter by switching to 64-bit or using a larger epsilon. Actually I find the result quite interesting: you can achieve reasonable accuracy with epsilon=1e-5.

renormalizer/mps/mps.py
```diff
@@ -696,11 +696,7 @@ def _evolve_dmrg_tdvp_vmf(self, mpo, evolve_dt) -> "Mps":
         else:
             coef = 1j
 
-        if not self.check_left_canonical():
-            self.move_qnidx(0)
-            self.left = True
-            self.canonicalise()
-            assert self.check_left_canonical()
+        self.ensure_left_canon()
```
jjren: Is there any API for a user to change the default atol in checking the canonicalization when using VMF?

liwt31: Not yet but I've modified some code in this commit to make this easier. Regarding the check-canonical issue, I think the best option is to figure out why using atol=1e-3 in the first place and restore it to the default value.

renormalizer/mps/tests/test_evolve.py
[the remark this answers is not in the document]
jjren: It's possible to pass the test with epsilon=1e-5. I think it is due to that the required accuracy is not that high (1e-2).
Ok, let's leave it here until we solve issue #25 .

renormalizer/mps/mps.py
jjren: Actually I don't know. You told me that some tests will fail if atol is set to the default value.

liwt31: Indeed... I'll pay attention to these tests.

PR #43, opened 2019-10-12
implement eom when considering overlap between SPF

jjren:
In the VMF/CMF evolution scheme, the gauge condition is fixed to be left-canonical. However, in the actual numerical calculation, the condition is not fulfilled because of the finite time step. If the time step size is not appropriate, the overlap between different SPF function should be considered <\Psi_i|\Psi_j> = S_ij \neq \delta_ij.

A flag `force_Ovlp=False`  in `EvolveConfig` controls this feature. 
True/False: w/o considering the Overlap Matrix. The default is `False` now.

### DIFF
--- a/renormalizer/mps/backend.py
+++ b/renormalizer/mps/backend.py
@@ -58,8 +58,8 @@ class Backend:
         self.first_mp = False
         self._real_dtype = None
         self._complex_dtype = None
-        self.use_32bits()
-        #self.use_64bits()
+        #self.use_32bits()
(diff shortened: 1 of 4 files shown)


jjren:
The annoying 32bits problem comes again. I will do this job (issue #25) these days. @liwt31

liwt31:
> The annoying 32bits problem comes again. I will do this job (issue #25) these days. @liwt31

I hope you find some of my comments in #25 helpful

PR #27, opened 2019-09-02
Fix scheme4

liwt31:
In scheme 4 all electron DOFs are in the same site.

Previously there's no zero exciton space in scheme4, which leads to a number of inconveniences.

This PR:
- add more docs (comments) on the scheme4 MPO
- add zero exciton state and QN for scheme4. All workarounds such as  `VirtualOnSite` can be removed

PR #47, opened 2019-11-07
improve charge transport

liwt31:
Add documents to `ChargeTransport` and add a new attribute -- k-space occupations. Details can be found in the code and documentation.

Some other changes:
- Refactored some code, including renaming the ambiguous `left` to `to_right` in order to be more explicit.
- Fixed a bug of not calculating reduced density matrices correctly in scheme 4, which is introduced in #27 .

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,12 +9,8 @@ install:
 
 script:
     - pytest --durations=0
-    - cd example; bash run.sh
-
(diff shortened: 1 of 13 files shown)


liwt31:
I forgot to included one of the doc files and the document for charge transport is not generated as expected. I'll fix this in the next commit.

PR #49, opened 2019-11-08
add finite temperature cv

jiangtong1000:
- Add finite temperature **correction vector** code 
- Modify zero temperature code by utilizing a super class for both zero/finite temperature conditions
Apart from this, *since correction vector at finite temperature is a MPO,* which also requires a initialization function `finiteT_cv` (implemented in `mps/mpo.py`). Besides, due to the algorithm similarity between ground state calculation and correction vector approach, a `select_Xbasis` is implemented in `mps/lib.py`, as the `select_basis` for MPS in ground state calculation


renormalizer/cv/tests/.ipynb_checkpoints/Untitled-checkpoint.ipynb
```diff
@@ -0,0 +1,6 @@
+{
```
liwt31: redundant file?

renormalizer/cv/tests/test_abs.py
```diff
@@ -0,0 +1,46 @@
...
+from renormalizer.cv.finitet import SpectraFtCV
+import numpy as np
+from renormalizer.tests.parameter import mol_list
+import os
+from renormalizer.cv.tests import cur_dir
+from renormalizer.utils import Quantity, CompressConfig
+import pytest
+
+
+def test_zt_abs(switch_to_64backend):
```
liwt31: `switch_to_64backend` is no longer available on master and tests are all run in 64-bit backend.

renormalizer/mps/mpo.py
```diff
@@ -547,6 +551,94 @@ def ph_onsite(cls, mol_list: MolList, opera: str, mol_idx:int, ph_idx=0):
...
+            x, xdim, xqn, compx = update_cv(u_set, s_set, qnset, None, nexciton, m_max, spectratype, percent=percent)
+            dim_list.append(xdim)
+            X.qn[ix + 1] = xqn
+            x = x.reshape(dim_list[-2], mol_list.pbond_list[ix], mol_list.pbond_list[ix], dim_list[ix + 1])
+            X.append(x)
+        dim_list.append(1)
+        X.append(np.random.random([dim_list[-2], mol_list.pbond_list[-1],
+                                  mol_list.pbond_list[-1], dim_list[-1]]))
+        X.qnidx = len(X) - 1
+        X.left = False
```
liwt31: Another piece of bad news... In master `left` is renamed to `to_right` in order to be more explicit in #47

liwt31:
Thank you for your work but this commit seems to have several conflicts with the master branch. My suggestion is to rebase to master locally, then solve the conflicts and in the end force-push.

renormalizer/mps/matrix.py
```diff
@@ -46,7 +46,10 @@ def __init__(self, array, dtype=None, is_full_mpdm=False):
         backend.running = True
 
     def __getattr__(self, item):
-        res = getattr(self.array, item)
+        try:
```
liwt31: Could you explain why this necessary?

jiangtong1000: It is noticed that in the `__init__` function inside `class matrix` exist `self.array`, however, once `self` has no attribute `array`, it calls `__getattr__`, inside which undergoes `getattr(self.array, item)`, Which brings to an endless loop between them, and finally lead to a maximum recursion number. I met such problems when attempting to enable *multiprocessing*, upon which I have discussed with you before.

jiangtong1000:
> Thank you for your work but this commit seems to have several conflicts with the master branch. My suggestion is to rebase to master locally, then solve the conflicts and in the end force-push.

Done~ Thanks for your kind help! I think it will be fine, let me know if can be better

renormalizer/mps/matrix.py
liwt31: Forgive me but I still can't understand this. I remember the discussion but not the details. I've got two questions:
- When will an instance of `Matrix` be accessed without the attribute `array`. Is it during multiprocessing? 
- How does the current code fix this? Will a maximum recursion error be raised and caught by the `except` clause? Leading to an `AttributeError`?

jiangtong1000: - Yes, such circumstance occurs only when using multiprocessing. I added `print(item)` after `except`, and I got a lot of `array` printed.
- I refered to [https://www.cnblogs.com/blackmatrix/p/5681480.html], then I found it worked! Still, I need further study to understand it.

jjren:
Hi Tong @jiangtong1000 ,

Could you please add the exact expression of what your code actually calculates in the comment section of ft-cv and zt-cv so that the user could use it more properly? (The constant prefactor should be take care of.)

jiangtong1000:
> Hi Tong @jiangtong1000 ,
> 
> Could you please add the exact expression of what your code actually calculates in the comment section of ft-cv and zt-cv so that the user could use it more properly? (The constant prefactor should be take care of.)

Sure, good suggestion! @jjren

liwt31:
According to the [CI](https://travis-ci.org/shuaigroup/Renormalizer/builds/610130214) the finite temperature test costs more than 1 min:
```
65.93s call     renormalizer/cv/tests/test_abs.py::test_ft_abs
```
Currently, we're trying to avoid time-consuming tests (#25 ). Is it possible to reduce the time cost?
Possible approaches:
- Use smaller system
- Calculate less data points

renormalizer/mps/matrix.py
liwt31: Here's what happens. When starting multiprocessing empty `Matrix`s are *pickled*, leading to the infinite recursion. Your clause will catch the maximum recursion error and turn it into an `AttributeError`. The error is propagated and caught by the `pickle` module which feels OK with `AttributeError`. So everything is working as expected.

I've improved your approach in the lastest commit in two ways:
- More documents on what happens.
- Raise `AttributeError` directly instead of finishing 500 times of recursion firstly. This should be beneficial to the performance.

jiangtong1000:
> According to the [CI](https://travis-ci.org/shuaigroup/Renormalizer/builds/610130214) the finite temperature test costs more than 1 min:
> 
> ```
> 65.93s call     renormalizer/cv/tests/test_abs.py::test_ft_abs
> ```
> 
> Currently, we're trying to avoid time-consuming tests (#25 ). Is it possible to reduce the time cost?
> Possible approaches:
> 
> * Use smaller system
> * Calculate less data points

now reduce to 16 second in local test

liwt31:
> > According to the [CI](https://travis-ci.org/shuaigroup/Renormalizer/builds/610130214) the finite temperature test costs more than 1 min:
> > ```
> > 65.93s call     renormalizer/cv/tests/test_abs.py::test_ft_abs
> > ```
> > 
> > 
> > Currently, we're trying to avoid time-consuming tests (#25 ). Is it possible to reduce the time cost?
> > Possible approaches:
> > 
> > * Use smaller system
> > * Calculate less data points
> 
> now reduce to 16 second in local test

Good to me but your force-push has erased my commit... Luckily the content is not much, but next time be careful when force-push.

PR #54, opened 2019-11-11
optmize charge transport and add lots of doc

liwt31:
Optimize `ChargeTransport` in the following ways:
- Optimize how physical observables are calculated now that things like `e_occupations` are not cached properties.
- Allow auto-load from previous calculated thermal propagated states (although they're exact).
- Skip processing `MpDm` in `ThermalProp` if exact because it's usually not necessary and sometimes time-consuming compared to the propagation time.
- `expand_bond_dimension` is also improved
    - check the dimension of `expander` during every iteration to ensure that the expanded MPS has the bond dimension required.
    - introduce states in excited states in case when the initial state is localized.

Also, add lots of doc to `ThermalProp`.


renormalizer/mps/mps.py
```diff
@@ -181,11 +181,15 @@ def random(cls, mol_list: MolList, nexciton, m_max, percent=1.0) -> "Mps":
...
     @classmethod
     def gs(cls, mol_list: MolList, max_entangled: bool):
-        """
-        T = \\infty maximum entangled GS state
-        electronic site: pbond 0 element 1.0
-                         pbond 1 element 0.0
-        phonon site: digonal element sqrt(pbond) for normalization
+        r"""
+        Obtain ground state at :math:`T = 0` or :math:`T = \infty` (maximum entangled).
+        Electronic DOFs are always at ground state. and vibronic DOFs depend on ``max_entangled``.
```
jjren: vibrational not vibronic

renormalizer/mps/mps.py
```diff
@@ -181,11 +181,15 @@ def random(cls, mol_list: MolList, nexciton, m_max, percent=1.0) -> "Mps":
...
+        For Spin-Boson model the electronic DOF also depends on ``max_entangled``.
+
+        Args:
+            mol_list (:class:`~renormalizer.model.MolList`): system information.
+            max_entanggled (bool): temperature of the vibronic DOFs. If set to ``True``,
```
jjren: The same as the last comment

renormalizer/mps/mps.py
jjren: @liwt31  You can fix it in the next commit.

liwt31: OK I'll keep this in mind

Issue #53, opened 2019-11-11
modify ft cv emission part


jiangtong1000:
- finite temperature emission still need further modification, cooperating with different imaginary evolution configs
- enable dumping density operator for further analysis.
- add some test for emission

PR #55, opened 2019-11-12
modify imaginary time evolution for emission calculation

jiangtong1000:
- modify imaginary time evolution for emission calculation, can dump density matrix for further analysis #53

### DIFF
--- a/renormalizer/cv/finitet.py
+++ b/renormalizer/cv/finitet.py
@@ -12,7 +12,12 @@ from renormalizer.mps.matrix import (
 from renormalizer.cv.spectra_cv import SpectraCv
 from renormalizer.mps import (Mpo, svd_qn, MpDm, ThermalProp)
 from renormalizer.mps.lib import update_cv
+from renormalizer.utils import (
+    CompressConfig, EvolveConfig,
+    CompressCriteria, EvolveMethod
+)
 import copy
+import os
 import logging
 import numpy as np
 import scipy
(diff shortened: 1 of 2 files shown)


liwt31:
This is a clever implementation but could be improved in several ways:
- It seems `period` function does not support scheme 4. Actually in scheme 4 where all electronic DOF is in the same site a much different approach should be used for periodic boundary condition MPO.
- If scheme 4 is taken into consideration, it'll be natural to make `period` an option of `Mpo`. Another reason for this is that other calculations may need to run in periodic conditions either.
- Adding some tests about the validity of the MPO should be nice.

Also a side note: it is not necessary to use `super()` for subclass methods every time. Do this only when the subclass has overwritten the ancestor's method and the overwritten method is desired.

jiangtong1000:
> This is a clever implementation but could be improved in several ways:
> 
> * It seems `period` function does not support scheme 4. Actually in scheme 4 where all electronic DOF is in the same site a much different approach should be used for periodic boundary condition MPO.
> * If scheme 4 is taken into consideration, it'll be natural to make `period` an option of `Mpo`. Another reason for this is that other calculations may need to run in periodic conditions either.
> * Adding some tests about the validity of the MPO should be nice.
> 
> Also a side note: it is not necessary to use `super()` for subclass methods every time. Do this only when the subclass has overwritten the ancestor's method and the overwritten method is desired.

OK, I will come back later after I understand scheme 4, and for now maybe jiajun @jjren  will do some tests for the new constructed function.

jjren:
I will write a more general MPO with periodic boundary condition. @jiangtong1000 @liwt31

liwt31:
> I will write a more general MPO with periodic boundary condition.

My suggestion is to set `periodic` as an attribute of `MolList` and construct MPO accordingly.

Issue #57, opened 2019-11-12
Do not use energy conservation as a criteria in imaginary time evolution


jjren:
https://github.com/shuaigroup/Renormalizer/blob/e59dd5d37b4b4f887d77f5a2f9581f07084b586e/renormalizer/mps/mps.py#L563-L575

In the imaginary time evolution, energy conservation should not be adopted as a criterion to adapt the time step size in both P&C and TDVP, I will fix this and modify the adaptive algorithm.
In my opinion, the wavefunction only is enough to determine the adaptive time step, because if the wavefunction has error O(\epsilon), the total energy has error O(\epsilon) too.

jjren:
Done.

PR #60, opened 2019-11-19
Adaptive dt

jjren:
There are several updates in this PR.
1. The adaptive P&C-RK45 and TDVP-PS TDVP_MU_CMF algorithms are re-designed (issue #57 ). All the available time evolution schemes: tdvp_vmf, tdvp_mu_vmf, tdvp_mu_cmf, tdvp_ps, P&C are now ready for both real time and imaginary time evolutions. In addition, all of them could be used in an adaptive mode. 
2. The default canonical condition of MPS is set to be rtol=1e-5, atol=1e-8 as `numpy.allclose`.
3. In class `Mpo`, a method `e_intersite` is added to create operator like a_j^\dagger a_i (Then the periodic boundary condition is easy in MPO scheme2).


renormalizer/mps/tests/test_mpdm.py
```diff
@@ -43,38 +43,51 @@ def test_from_mps():
 @pytest.mark.parametrize(
     "mol_list, etot_std, occ_std, rtol",
     (
-        [
-            parameter.hybrid_mol_list,
-            0.0853441664951,
-            [0.20881609, 0.35239430, 0.43878960],
-            5e-3,
-        ],
+        # explicit hybrid algorithm will be removed
```
liwt31: So the tests for the hybrid scheme currently do not work?

renormalizer/mps/tests/test_mpdm.py
```diff
@@ -43,38 +43,51 @@ def test_from_mps():
...
     mpo = Mpo(mol_list)
     beta = Quantity(298, "K").to_beta()
     evolve_time = beta / 2j
-    if nsteps is None:
-        evolve_config = EvolveConfig(evolve_method, adaptive=True, evolve_dt=beta/100j)
+    
+    evolve_config = EvolveConfig(evolve_method, adaptive=adaptive,
+                guess_dt=0.1/1j)
+    
+    if adaptive == True:
```
liwt31: BTW, usually `if adaptive:` is OK.

renormalizer/mps/mpo.py
```diff
@@ -554,6 +558,69 @@ def ph_onsite(cls, mol_list: MolList, opera: str, mol_idx:int, ph_idx=0):
         mpo.build_empty_qn()
         return mpo
 
+    @classmethod
```
liwt31: Still, I don't think this method works for scheme 4. If it is currently not necessary at least we can check the `mol_list` to ensure scheme 4 is not used.

renormalizer/mps/tests/test_mpo.py
```diff
@@ -75,6 +75,34 @@ def test_scheme4():
...
+    
+    mpo5 = Mpo.e_intersite(mol_list, {2:r"a^\dagger a"}, Quantity(0.5))
+    assert mpo5.add(mpo5).distance(mpo4) == pytest.approx(0, abs=1e-5)
+    
+    mpo6 = Mpo.e_intersite(mol_list, {0:r"a^\dagger",2:"a"}, Quantity(1.0))
+    mpo7 = Mpo.onsite(mol_list, "a", mol_idx_set=[2])
+    assert mpo2.apply(mpo7).distance(mpo6) == pytest.approx(0, abs=1e-5)
+    
+    mpo8 = Mpo(mol_list)
+    mol_list1 = mol_list.switch_scheme(2)
```
liwt31:
```suggestion
    # a dirty hack to switch from scheme 2 to scheme 3
    mol_list1 = mol_list.switch_scheme(2)
```

liwt31:
According to Travis CI this PR seems to cause a noticeable performance regression.
[Most time-consuming tests in master](https://travis-ci.org/shuaigroup/Renormalizer):
```
110.02s call     renormalizer/mps/tests/test_quasiboson.py::test_quasiboson_solver[value0]
76.40s call     renormalizer/spectra/tests/test_hybrid_ft_spectra.py::test_ft_hybrid_dmrg_tdh[mol_list0-abs-hybrid_FT_abs_pure.npy]
57.87s call     renormalizer/spectra/tests/test_hybrid_spectra.py::test_hybrid_emi[1-mol_list1-ZeroExactEmi.npy-0.01]
57.60s call     renormalizer/spectra/tests/test_hybrid_spectra.py::test_hybrid_emi[2-mol_list3-ZeroExactEmi.npy-0.01]
```
[While in this PR](https://travis-ci.org/shuaigroup/Renormalizer/builds/614331331):
```
126.82s call     renormalizer/spectra/tests/test_hybrid_ft_spectra.py::test_ft_hybrid_dmrg_tdh[mol_list2-emi-hybrid_FT_emi_pure.npy]
118.49s call     renormalizer/spectra/tests/test_hybrid_ft_spectra.py::test_ft_hybrid_dmrg_tdh[mol_list0-abs-hybrid_FT_abs_pure.npy]
114.25s call     renormalizer/transport/tests/test_hybrid_dynamics.py::test_FT_dynamics_hybrid_TDDMRG_TDH[10-4]
103.68s call     renormalizer/transport/tests/test_hybrid_dynamics.py::test_FT_dynamics_hybrid_TDDMRG_TDH[10-3]
99.97s call     renormalizer/mps/tests/test_quasiboson.py::test_quasiboson_solver[value0]
```
Any suggestion on the possible causes?
I also noticed that the criteria for tests in ` renormalizer/spectra/tests/test_hybrid_ft_spectra.py ` are loosened. I suppose there should be a particular reason for the behavior.
I'll look into this recently when I have some time.

renormalizer/mps/mps.py
```diff
@@ -1566,3 +1622,15 @@ def __str__(self):
...
     def __iter__(self):
         return iter((self.bra_mps, self.ket_mps))
+
+
+def min_abs(t1, t2):
+    # t1, t2 could be int, float, complex
+    # return the number with smaller norm
+
+    if xp.iscomplex(t1) != xp.iscomplex(t2):
+        raise TypeError(f"{xp.iscomplex(t1)} != {xp.iscomplex(t2)}")
```
liwt31: I suspect the error message would be some like `True != False` which is not very helpful.

renormalizer/mps/tests/test_mpdm.py
jjren: It works. I skip it to save time.

renormalizer/mps/mpo.py
liwt31: And hopefully this method can take the place of `displacement`. I think I can handle this in the near future.

renormalizer/mps/mps.py
```diff
@@ -641,17 +650,19 @@ def _evolve_dmrg_tdvp_vmf(self, mpo, evolve_dt) -> "Mps":
         else:
             coef = 1j
         
-        self.ensure_left_canon()
+        if not (self.evolve_config.force_ovlp == True and self.to_right == False):
```
liwt31:
```suggestion
        if not (self.evolve_config.force_ovlp and not self.to_right):
```

renormalizer/mps/mps.py
```diff
@@ -748,6 +762,13 @@ def func_vmf(t,y):
             offset += mps[imps].size
 
         logger.debug(f"{self.evolve_config.method} VMF func called: {sol.nfev}. RKF steps: {len(sol.t)}")
+        
+        # switch to tdvp_mu_vmf
+        w_min_list = xp.array(w_min_list)
+        logger.debug(f"w.min={w_min_list.min()}, Switch to tdvp_mu_vmf")
```
liwt31:
```suggestion
```

renormalizer/mps/mpo.py
jjren: What is displacement for ?

jjren:
> According to Travis CI this PR seems to cause a noticeable performance regression.
> [Most time-consuming tests in master](https://travis-ci.org/shuaigroup/Renormalizer):
> 
> ```
> 110.02s call     renormalizer/mps/tests/test_quasiboson.py::test_quasiboson_solver[value0]
> 76.40s call     renormalizer/spectra/tests/test_hybrid_ft_spectra.py::test_ft_hybrid_dmrg_tdh[mol_list0-abs-hybrid_FT_abs_pure.npy]
> 57.87s call     renormalizer/spectra/tests/test_hybrid_spectra.py::test_hybrid_emi[1-mol_list1-ZeroExactEmi.npy-0.01]
> 57.60s call     renormalizer/spectra/tests/test_hybrid_spectra.py::test_hybrid_emi[2-mol_list3-ZeroExactEmi.npy-0.01]
> ```
> 
> [While in this PR](https://travis-ci.org/shuaigroup/Renormalizer/builds/614331331):
> 
> ```
> 126.82s call     renormalizer/spectra/tests/test_hybrid_ft_spectra.py::test_ft_hybrid_dmrg_tdh[mol_list2-emi-hybrid_FT_emi_pure.npy]
> 118.49s call     renormalizer/spectra/tests/test_hybrid_ft_spectra.py::test_ft_hybrid_dmrg_tdh[mol_list0-abs-hybrid_FT_abs_pure.npy]
> 114.25s call     renormalizer/transport/tests/test_hybrid_dynamics.py::test_FT_dynamics_hybrid_TDDMRG_TDH[10-4]
> 103.68s call     renormalizer/transport/tests/test_hybrid_dynamics.py::test_FT_dynamics_hybrid_TDDMRG_TDH[10-3]
> 99.97s call     renormalizer/mps/tests/test_quasiboson.py::test_quasiboson_solver[value0]
> ```
> 
> Any suggestion on the possible causes?
> I also noticed that the criteria for tests in `renormalizer/spectra/tests/test_hybrid_ft_spectra.py` are loosened. I suppose there should be a particular reason for the behavior.
> I'll look into this recently when I have some time.

renormalizer/mps/mpo.py
liwt31: Similar to this method but somehow limited. Used when creating flux operators.

jjren:
> > According to Travis CI this PR seems to cause a noticeable performance regression.
> > [Most time-consuming tests in master](https://travis-ci.org/shuaigroup/Renormalizer):
> > ```
> > 110.02s call     renormalizer/mps/tests/test_quasiboson.py::test_quasiboson_solver[value0]
> > 76.40s call     renormalizer/spectra/tests/test_hybrid_ft_spectra.py::test_ft_hybrid_dmrg_tdh[mol_list0-abs-hybrid_FT_abs_pure.npy]
> > 57.87s call     renormalizer/spectra/tests/test_hybrid_spectra.py::test_hybrid_emi[1-mol_list1-ZeroExactEmi.npy-0.01]
> > 57.60s call     renormalizer/spectra/tests/test_hybrid_spectra.py::test_hybrid_emi[2-mol_list3-ZeroExactEmi.npy-0.01]
> > ```
> > 
> > 
> > [While in this PR](https://travis-ci.org/shuaigroup/Renormalizer/builds/614331331):
> > ```
> > 126.82s call     renormalizer/spectra/tests/test_hybrid_ft_spectra.py::test_ft_hybrid_dmrg_tdh[mol_list2-emi-hybrid_FT_emi_pure.npy]
> > 118.49s call     renormalizer/spectra/tests/test_hybrid_ft_spectra.py::test_ft_hybrid_dmrg_tdh[mol_list0-abs-hybrid_FT_abs_pure.npy]
> > 114.25s call     renormalizer/transport/tests/test_hybrid_dynamics.py::test_FT_dynamics_hybrid_TDDMRG_TDH[10-4]
> > 103.68s call     renormalizer/transport/tests/test_hybrid_dynamics.py::test_FT_dynamics_hybrid_TDDMRG_TDH[10-3]
> > 99.97s call     renormalizer/mps/tests/test_quasiboson.py::test_quasiboson_solver[value0]
> > ```
> > 
> > 
> > Any suggestion on the possible causes?
> > I also noticed that the criteria for tests in `renormalizer/spectra/tests/test_hybrid_ft_spectra.py` are loosened. I suppose there should be a particular reason for the behavior.
> > I'll look into this recently when I have some time.

It is due to that in P&C, the compression criteria are changed (The magic +4 is removed). Therefore the result is changed a little bit. I've checked the correlation function, it's correct but a little bit different from the reference.

renormalizer/mps/mpo.py
jjren: I see.

liwt31:
>It is due to that in P&C, the compression criteria are changed (The magic +4 is removed). Therefore the result is changed a little bit. I've checked the correlation function, it's correct but a little bit different from the reference.

Aha so the magic does the trick then. Not bad for two lines of code 😆 . May be for better performance the tests using PC should be switched to PS or VMF in the future.

jjren:
Yes. I think so, since the tdvp_vmf and tdvp_ps methods are more robust to use now.

renormalizer/mps/mpo.py
jjren: I add a `mol_list` check in this function. Please see the new update.

renormalizer/mps/mps.py
jjren: I've fixed it. Please see the new update.

liwt31:
I noticed that vmf methods are optimized so that we can automatically switch between using MU and not using MU. This is great because VMF without MU is much cheaper on GPU. 
However in this case I think we should only provide one option in `EvolveMethod`. Otherwise maybe another flag in `EvolveConfig` is required.

jjren:
> I noticed that vmf methods are optimized so that we can automatically switch between using MU and not using MU. This is great because VMF without MU is much cheaper on GPU.
> However in this case I think we should only provide one option in `EvolveMethod`. Otherwise maybe another flag in `EvolveConfig` is required.

This is what I thought. But I think we still need some time to see if this auto-switch really works or not. After that, we can safely remove the tdvp_vmf interface and only provides tdvp_mu_vmf.

liwt31:
I'm OK with this PR now, only that the commit history is a bit messy. I'm going to squash some of the commits into one and then force push.

PR #63, opened 2019-11-22
Optimize vmf

jjren:
The main change in this PR is that
1. tdvp_mu_vmf and tdvp_vmf are combined into a single function and the integrator RK45 is carried out with the constraint of quantum number symmetry.
2. the default `force_ovlp` in class `EvolveConfig` is set to be True.


liwt31:
This seems to be a great commit. Thanks.

PR #66, opened 2019-11-23
Drastic refactor backend to save GPU memory

liwt31:
Previously in GPU backend we store all matrices in GPU and copy to host when necessary to achieve maximum speed. However, it turns out in GPU calculation total memory usage is the bottle-neck instead of host-device memory transfer.
In this PR I make a drastic change to the strategy: **matrices are always stored in host and are copied to GPU when necessary.**
The new strategy allows incredibly small memory usage when scheme <= 3 ( save up to 80% memory according to some tests) and has little effect on scheme4, in which the memory bottleneck is not storing MPSs but contraction intermediate results. The performance regression is well controlled, typically less than 20%, maybe more serious in VMF.
The new strategy also encourages the developers to use numpy or cupy ndarray directly instead of the `Matrix` class, which only stores matrices in host memory. Previously we usually use `Matrix(a)` to transfer matrix `a` to GPU memory. Now the correct way is `asxp(a)`. As a result, the developers should be ultra-careful regarding SP/DP because previously this is guaranteed in the `Matrix` class and now it is the responsibility of developers. For example, one should use things like
```
a = xp.ones((1,1,1), dtype=backend.real_dtype)
```
instead of 
```
a = xp.ones((1,1,1))
```
and the old way is
```
a = ones((1,1,1))
```

I've also updated codes in `cv` and they can work in CPU but they are not tested on GPU due to some unknown error. @jiangtong1000 you might want to know about the changes and review the cv code.

### DIFF
--- a/renormalizer/mps/matrix.py
+++ b/renormalizer/mps/matrix.py
@@ -7,8 +7,7 @@ from typing import List, Union
 
 import numpy as np
 
-from renormalizer.mps.backend import xp, backend, cp
-
+from renormalizer.mps.backend import np, backend, xp, USE_GPU, xp as xp
 
 logger = logging.getLogger(__name__)
 
@@ -31,15 +30,16 @@ class Matrix:
 
     def __init__(self, array, dtype=None, is_full_mpdm=False):
         assert array is not None
+        array = asnumpy(array)
         if dtype == backend.real_dtype:
             # forbid unchecked casting
-            assert not xp.iscomplexobj(array)
+            assert not np.iscomplexobj(array)
         if dtype is None:
-            if xp.iscomplexobj(array):
+            if np.iscomplexobj(array):
                 dtype = backend.complex_dtype
             else:
                 dtype = backend.real_dtype
-        self.array: xp.ndarray = xp.asarray(array, dtype=dtype)
+        self.array: np.ndarray = np.asarray(array, dtype=dtype)
         self.original_shape = self.array.shape
         self.sigmaqn = None
         self.is_full_mpdm = is_full_mpdm
@@ -50,14 +50,14 @@ class Matrix:
         # see https://stackoverflow.com/questions/22781872/python-pickle-got-acycle-recursion-with-getattr
         array = super().__getattribute__("array")
         res = getattr(array, item)
-        if isinstance(res, xp.ndarray):
+        if isinstance(res, np.ndarray):
             return Matrix(res)
         functiontype = type([].append)
         if isinstance(res, functiontype):
 
             def wrapped(*args, **kwargs):
                 res2 = res(*args, **kwargs)
-                if isinstance(res2, xp.ndarray):
+                if isinstance(res2, np.ndarray):
                     return Matrix(res2)
                 return res2
 
@@ -71,20 +71,14 @@ class Matrix:
 
     def astype(self, dtype):
         assert not (self.dtype == backend.complex_dtype and dtype == backend.real_dtype)
-        self.array = xp.asarray(self.array, dtype=dtype)
+        self.array = np.asarray(self.array, dtype=dtype)
         return self
 
     def abs(self):
-        return self.__class__(xp.abs(self.array))
+        return self.__class__(np.abs(self.array))
 
     def norm(self):
-        return xp.linalg.norm(self.array.flatten())
-
-    def asnumpy(self):
-        if xp == np:
-            return self.array
-        else:
-            return xp.asnumpy(self.array)
+        return np.linalg.norm(self.array.flatten())
 
     # physical indices exclude first and last indices
     @property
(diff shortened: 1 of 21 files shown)


jjren:
It‘s pretty good! The memory should be the bottleneck when dealing with large systems.
Since this PR is a rather big change, I think it is better to give some development instructions after the next group meeting.

liwt31:
Thanks for your correction @jiangtong1000 . I'll surely talk about this in the next group meeting

jiangtong1000:
@liwt31 I made some small bugfixs based on your pr, maybe the many commit messages can be cleaned up when this pr is ready to merge

liwt31:
I think we need a quick decision on this PR, because the probability of conflicts is high.

jjren:
Yes. We should discuss tomorrow. maybe afternoon.

renormalizer/mps/matrix.py
```diff
@@ -117,26 +111,26 @@ def check_lortho(self, rtol=1e-5, atol=1e-8):
...
     def check_rortho(self, rtol=1e-5, atol=1e-8):
         """
         check R-orthogonal
         """
-        tensm = self.array.reshape([self.shape[0], np.prod(self.shape[1:])])
-        s = xp.dot(tensm, tensm.T.conj())
-        return allclose(s, xp.eye(s.shape[0]), rtol=rtol, atol=atol)
+        tensm = asxp(self.array.reshape([self.shape[0], np.prod(self.shape[1:])]))
+        s = tensm @ tensm.T.conj()
+        return xp.allclose(s, np.eye(s.shape[0]), rtol=rtol, atol=atol)
```
jjren: Should this `np.eye` be replaced with `xp.eye`?

renormalizer/mps/matrix.py
```diff
@@ -7,8 +7,7 @@
 
 import numpy as np
 
-from renormalizer.mps.backend import xp, backend, cp
-
+from renormalizer.mps.backend import np, backend, xp, USE_GPU, xp as xp
```
jjren: ???

jiangtong1000:
@ I made some changes based on this pr to enable CV also practical for GPU, however it seems that code changes only happen on the branch `save_gpu_memory` of your forked `Renormalizer`

jiangtong1000:
@liwt31  I made some changes based on this pr to enable CV also practical for GPU, however it seems that code changes only happen on the branch `save_gpu_memory` of your forked `Renormalizer`

jiangtong1000:
> @liwt31 I made some changes based on this pr to enable CV also practical for GPU, however it seems that code changes only happen on the branch `save_gpu_memory` of your forked `Renormalizer`

jiangtong1000:
> @liwt31 I made some changes based on this pr to enable CV also practical for GPU, however it seems that code changes only happen on the branch `save_gpu_memory` of your forked `Renormalizer`

Forgive me, There may be some delay

jiangtong1000:
Why this pr hangs here for so long?😂

renormalizer/mps/matrix.py
liwt31: I think this is a bug of the refactoring tool of PyCharm...

renormalizer/mps/matrix.py
liwt31: makes sense to me...

liwt31:
> Why this pr hangs here for so long?😂

Yeah I think it's about time addressing this. I'll take a look tomorrow and see if we can get this done before the weekend.

liwt31:
> @liwt31 I made some changes based on this pr to enable CV also practical for GPU, however it seems that code changes only happen on the branch `save_gpu_memory` of your forked `Renormalizer`

I've resolved the conflicts between this branch and the master branch, you may wish to rebase to the latest branch to resolve your conflicts inherited from this branch.

jiangtong1000:
> > @liwt31 I made some changes based on this pr to enable CV also practical for GPU, however it seems that code changes only happen on the branch `save_gpu_memory` of your forked `Renormalizer`
> 
> I've resolved the conflicts between this branch and the master branch, you may wish to rebase to the latest branch to resolve your conflicts inherited from this branch.

Currently I'm using code from master branch, so I'm fine with all these changes.

renormalizer/mps/mps.py
```diff
@@ -375,10 +372,10 @@ def expectation(self, mpo, self_conj=None) -> Union[float, complex]:
         else:
             ret_float = False
         environ = Environ(self, mpo, "R", mps_conj=self_conj)
-        l = ones((1, 1, 1))
+        l = xp.ones((1, 1, 1), dtype=self.dtype)
         r = environ.read("R", 1)
         path = self._expectation_path()
-        val = multi_tensor_contract(path, l, self[0], mpo[0], self_conj[0], r).array
+        val = multi_tensor_contract(path, l, self[0], mpo[0], self_conj[0], r)
         if ret_float:
```
jjren: The `ret_float` is dangerous, only when the mpo is hermitian, the expectation value is real.

renormalizer/mps/mps.py
```diff
@@ -1597,7 +1624,7 @@ def calc_ft(self):
             dot = self.bra_mps.conj().dot(self.ket_mps)
         else:
             dot = self.bra_mps.conj().expectation(self.mpo, self.ket_mps)
-        return (
+        return complex(
```
jjren: I don't fully understand why we should transfer the numpy array to python float/complex explicitly?
I found such a transformation in other places.

renormalizer/mps/matrix.py
```diff
@@ -7,8 +7,7 @@
 
 import numpy as np
```
jjren: import numpy twice here

renormalizer/mps/mps.py
liwt31: Currently, we only have hermitian operators. If a non-hermitian operator is used it is a significant change and the user would expect complex output so that the bug can be found without much effort. The reason why it is designed this way is that it avoids annoying warnings issued by `matplotlib` when ploting imaginary numbers. I agree to add an assertion here though.

renormalizer/mps/mps.py
liwt31: There are two reasons for the transformation.
- The first is related to the JSON output. Numpy float can not be serialized by JSON and has to be converted to Python float first.
- The second is that if Cupy array is involved there could be unexpected behavior. For example
    ```
    a = cp.random.rand(1, 5)
    b = cp.random.rand(5, 1)
    # suppose we calculate the expectation value. Note type of `c` is Cupy float
    c = (a @ b).ravel()[0] 
    d = np.array([c, c]) # `d` become an array of array, not an array of number.
    ```
Converting results that are expected to be numbers to Python scalars can help solving the problem.

renormalizer/mps/matrix.py
liwt31: I'm going to fix this. Thanks.

PR #75, opened 2019-12-26
a new class Property

jjren:
1. In class `TdMpsJob`,  `dump_type` is added, which could dump results in `JSON`(`dump_type=".json"`) or `numpy npz`(`dump_type=".npz"`, default) format to solve that JSON does not support numpy array or complex number.

2. A new class `Property` allows user-defined MPO to be calculated during TD simulation or for one single state (refer to `thermalprop` or test).

3. `property/ops.py` contains operators for property calculation. 
Only the electron phonon static correlation function operator is available now.

### DIFF
--- a/renormalizer/mps/tests/test_evolve.py
+++ b/renormalizer/mps/tests/test_evolve.py
@@ -71,7 +71,9 @@ def test_tdvp_vmf(init_state, with_mu, force_ovlp, atol):
 def test_tdvp_cmf(init_state):
     mps = init_state.copy()
     mps.evolve_config  = EvolveConfig(EvolveMethod.tdvp_mu_cmf)
-    check_result(mps, mpo, 0.01, 0.5, 5e-4)
+    for flag in [True, False]:
+        mps.evolve_config.tdvp_cmf_c_trapz = flag
+        check_result(mps, mpo, 0.01, 0.5, 5e-4)
 
 
 @pytest.mark.parametrize("init_state", (
--- a/renormalizer/mps/tests/test_evolve.py
+++ b/renormalizer/mps/tests/test_evolve.py
@@ -80,7 +80,7 @@ def test_tdvp_cmf(init_state):
 def test_tdvp_ps(init_state):
     mps = init_state.copy()
     mps.evolve_config  = EvolveConfig(EvolveMethod.tdvp_ps)
-    check_result(mps, mpo, 0.5, 5)
+    check_result(mps, mpo, 0.4, 5)
 
 
 # used for debugging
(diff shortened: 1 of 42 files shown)


renormalizer/mps/tests/test_evolve.py
```diff
@@ -71,7 +71,9 @@ def test_tdvp_vmf(init_state, with_mu, force_ovlp, atol):
 def test_tdvp_cmf(init_state):
     mps = init_state.copy()
     mps.evolve_config  = EvolveConfig(EvolveMethod.tdvp_mu_cmf)
-    check_result(mps, mpo, 0.01, 0.5, 5e-4)
+    for flag in [True, False]:
```
liwt31: It'll be more consistent if the option is moved to the decorator `pytest.mark.parametrize`

renormalizer/property/ops.py
```diff
@@ -0,0 +1,64 @@
...
+
+import numpy as np
+
+
+def e_ph_static_correlation(mol_list: MolList, imol:int =0, jph:int =0,
+        periodic:bool =False, name:str="S"):
+    '''
+    construct the electron-phonon static correlation operator in polaron problem
+    if periodic:
+        S_(m, jph) = \frac{1}{D} \sum_n \langle x_{m+n,jph} a_n^\dagger a_n \rangle
```
liwt31: What is the definition of `D`? Is this implying periodic *homogeneous* model?

renormalizer/property/property.py
```diff
@@ -0,0 +1,66 @@
...
+            mps_conjs:Union[Dict[str, Mps], Dict[str, MpDm], Mps, MpDm, None] =None):
+        '''
+        calculate all the properties with same mps or mps_conj
+        or 
+        calculate each property with different {prop_str:mps}, {prop_str:mps_conj}
+        '''
+        for prop_str in self.prop_strs:
+            
+            # pick ket
+            if mpss.is_mps or mpss.is_mpdm:
```
liwt31: If `Dict[str, Mps]` is passed in as `mpss` then this line would result in an `AttributeError`.  I suggest deleting the logic for `Dict` and create a new method when the necessity comes.

jjren: Yes. It's a bug.

renormalizer/property/ops.py
jjren: D is the displacement. I will add a reference paper.

renormalizer/mps/tests/test_evolve.py
jjren: fixed.

renormalizer/property/property.py
jjren: fixed

PR #78, opened 2020-01-03
enable period h_mpo init

jiangtong1000:
Now we can assign the `mol_list` 's`period` whether to init mpo with boundary condition

### DIFF
--- a/renormalizer/cv/spectra_cv.py
+++ b/renormalizer/cv/spectra_cv.py
@@ -7,6 +7,7 @@ import multiprocessing
 from renormalizer.mps import Mpo
 from renormalizer.utils import Quantity
 from renormalizer.utils.elementop import construct_e_op_dict, ph_op_matrix
+import importlib.util
 import logging
 import time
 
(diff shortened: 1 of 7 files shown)


renormalizer/cv/spectra_cv.py
```diff
@@ -49,22 +50,20 @@ def check_qn(self, X):
         print('Quantum number of X', check_1)
     '''
     def run(self):
-        try:
+        if importlib.util.find_spec("cupy"):
```
liwt31: Could you explain why `forkserver` is necessary for `cupy`? I'm sorry but I'm not familiar with this feature.

jiangtong1000: I found the same problem and solution at [pytorch/pytorch/issues/334] and [chainer/chainer/issues/1087], my understanding is that before starting multiprocessing, the cuda has already been initialized, so actually each process did't inherent the cuda resource, that means cuda should be initialized at each process separately by setting the start method as `spwan` or `forkserver`

liwt31: Thanks for the explanation.

liwt31:
I'm OK this PR but as you've requested the review from jjren I'll leave this open for a while.

PR #79, opened 2020-01-04
optimize part of old heavy ft_cv code, speed up ~ 10%

jiangtong1000:
improve some code relevant to quantum number calculation in finite t cv

### DIFF
--- a/renormalizer/cv/finitet.py
+++ b/renormalizer/cv/finitet.py
@@ -3,7 +3,6 @@
 # finite temperature absorption/emission spectrum based on Correction vector
 
 from renormalizer.mps.matrix import (
-    Matrix,
     multi_tensor_contract,
     tensordot,
     moveaxis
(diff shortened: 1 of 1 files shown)


renormalizer/cv/finitet.py
```diff
@@ -429,86 +435,49 @@ def construct_X_qnmat(self, addlist, direction):
...
-        condition = []
-        for i in range(0, len(list_qnmat), 2):
-            if (list_qnmat[i] == 0) or (list_qnmat[i + 1] == 0):
-                condition.append(False)
-            else:
-                condition.append(True)
-        condition = np.array(condition)
+        condition = condition.all(axis=-1)
         condition = condition.reshape(mat_shape)
         return condition
```
jiangtong1000: maybe you can give me some advice on improving this function, constitutes nearly 5% of the whole calculation apart from the rest 95% from cg. This function is used to compare the qn matrix with the required `qn`, `qn` is a list, eg. [0, 1] describing the quantum number of a local site CV mpo

jjren: https://docs.scipy.org/doc/numpy/reference/arrays.dtypes.html

a = np.array([(1,1),(1,0)], dtype='i,i')
a == np.array((1,0),dtype='i,i')

liwt31: IMO it's not worthy of your time to pursue 5% possible speedup. If I were you I would rather play video games...

PR #81, opened 2020-01-21
add scheme 4 support for transport autocorr

liwt31:
- add method `copy` to `MolList`
- change the default behavior of `build empty qn` so that the `qnidx` is set to be the right-most site to be consistent with other normal MPS/MPO
- Add MPO `intersite` code for scheme 4 and modify related tests. Transport autocorr was unable to run in scheme 4 because of this.

### DIFF
--- a/renormalizer/model/mlist.py
+++ b/renormalizer/model/mlist.py
@@ -15,7 +15,7 @@ from renormalizer.utils.utils import cast_float
 class MolList:
 
     def __init__(self, mol_list: List[Mol], j_matrix: Union[Quantity, np.ndarray, None], scheme: int = 2, periodic: bool = False):
-        self.period = periodic
+        self.periodic = periodic
         self.mol_list: List[Mol] = mol_list
 
         # construct the electronic coupling matrix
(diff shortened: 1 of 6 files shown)


renormalizer/mps/mpo.py
```diff
@@ -586,45 +585,76 @@ def intersite(cls, mol_list: MolList, e_opera: dict, ph_opera: dict, scale:
...
+                        idx2, op2 = next(it)
+                        assert idx1 != idx2
+                        assert {op1, op2} == {r"a^\dagger", "a"}
+                        if op1 == "a":
+                            mo[idx2+1, idx1+1] = 1
+                        else:
+                            mo[idx1+1, idx2+1] = 1
+                    else:
+                        assert False
                     mpo.qn.append(mpo.qn[-1])
```
jjren: `qn` seems not correct if op == r"a^\dagger"?

liwt31:
Thanks for the review. I've fixed the qn bug and added a new option (`periodic`) for MolList

renormalizer/model/mlist.py
```diff
@@ -14,8 +14,8 @@
 
 class MolList:
 
-    def __init__(self, mol_list: List[Mol], j_matrix: Union[Quantity, np.ndarray, None], scheme: int=2, period=False):
-        self.period = period
+    def __init__(self, mol_list: List[Mol], j_matrix: Union[Quantity, np.ndarray, None], scheme: int = 2, periodic: bool = False):
+        self.period = periodic
```
jjren: It is better to replace all the `period` with `periodic`.

liwt31: I agree. But I think @jiangtong1000 may have some code that uses this property. @jiangtong1000 are you okay with this modification?

jiangtong1000: SURE

PR #82, opened 2020-02-18
General Symbolic MPO Construction Routine

jjren:
***a main update***

1. a general symbolic MPO construction routine `Mpo.general_mpo` is added. It is convenient to construct any operator already implemented (Holstein, SBM, intersite, onsite, flux operator...) or not yet implemented (Peierls, general vibronic model..) in the package now.

2. a simple new class `Op` is added to `utils.elementop` and to be the internal operator data structure used in the future. 

3. a simple new class `BasisSet` is added to `renormalizer/utils/basis.py`. And several derived classes are added for the simple two-level electronic state (scheme1/2/3), multi-level electronic state (scheme4),  SBM half-spin and simple harmonic oscillator (SHO). The `sigmaqn` is an inherent property of BasisSet and some old flags such as `ephtable`, `pbond_list` are redundant. First quantized operators such as `x`, `x^2`, `p`, `p^2` and second quantized operators such as `b^\dagger`, `b` are all supported on the vibrational basis sets.
 
4. combine the symbolic MPO in 1, Op in 2 and BasisSet in 3, an MPO in matrix representation could be obtained.

5. Since `MolList` is not compatible to the general model such as general vibronic model, a new MolList2 is added. The key parameter defined in MolList2 is the order of DoFs, BasisSet, Model Hamiltonian or Operator and a translator of the user-defined model. Two translators are implemented `vibronic_model` and `general_model`.

6. A bi-linear vibronic model (in `renormalizer/vibronic/test`) for Pyrazine with 4/24 modes has been tested.

***7**. This main update especially `General Symbolic MPO Construction Routine` is an experimental feature without comprehensive tests.  Please be careful to use it in a productive calculation. 
The main attention should be paid to the case that multi-electronic states are on a single site.
In the scheme4, a zero exciton state is added, so there are lots of places "+1" is required to get the parameter such as `J[i,j]`. However, the gs is not guaranteed to be added in MolList2, so the developer should be careful to reuse the existed methods designed for MolList scheme4.   

--------------------------2020.4.15---------------
1. the `_model_translator_general_model` and `symbolic_mpo` function are optimized.
2. ab initio QC Hamiltonian is tested.  `/example/h2o_qc.py`


jiangtong1000:
Impressive updates, however many modules above I am not so familiar with, so maybe I will spend some time learning it firstly.  Don't wait for me if you guys think everything works fine. 😅

liwt31:
I've modified some code as part of the understanding process. The modification has not touched the core logic of the algorithm yet and it's primarily maintenance work.
- Refactor `BasisSHO` in `BasisSet` to make it shorter.
- Refactor `test_pyr4.py`. Combining the two tests into one and change the standard data to MCTDH. I reckon the reason why it was not so is that `expand_bond_dimension` was not used and data was obtained via variable step-size.

Sorry to directly push to the jjren_dev branch. I hope it'll not be a big problem to have two people working on the same branch.

renormalizer/utils/basis.py
```diff
@@ -46,113 +47,90 @@ def op_mat(self, op):
...
+
         elif op_symbol == "p":
-            #<m|p|n> = -i sqrt(w/2) sqrt(n) delta(m,n-1) - sqrt(n+1) delta(m,n+1)
-            mat = np.zeros((self.nbas, self.nbas), dtype=np.complex128)
-            
-            for iket in range(1, self.nbas):
-                mat[iket-1, iket] += -1.0j * np.sqrt(0.5*self.omega) * np.sqrt(iket) 
-            for iket in range(self.nbas-1):
-                mat[iket+1, iket] -= -1.0j * np.sqrt(0.5*self.omega) * np.sqrt(iket+1)
+            # <m|p|n> = -i sqrt(w/2) <m| b^\dagger - b |n>
```
jjren: I think the second quantization of p is wrong. 
-i sqrt(w/2) <m| b - b^\dagger |n>

jjren:
> I reckon the reason why it was not so is that `expand_bond_dimension` was not used and data was obtained via variable step-size.

Actually, you've pointed out the reason "for multi-e case the `expand bond dimension` routine is currently not working". That's why I use P&C to warm up this simple test.

liwt31:
> > I reckon the reason why it was not so is that `expand_bond_dimension` was not used and data was obtained via variable step-size.
> 
> Actually, you've pointed out the reason "for multi-e case the `expand bond dimension` routine is currently not working". That's why I use P&C to warm up this simple test.

Apparently, the reason why the routine is not working is that multi-e basis has no creation/annihilation operator because the state with no electron is not represented.

I wonder why this is the case given that scheme 4 has already demonstrated the zero-electron state is necessary (#27).

Another side-note: in scheme 4 the basis is arranged as  |vacuum>, |1>...|m>, which leads to lots of `+1` to the code as mentioned in this PR. This could be avoided by rearranging the basis as |1>...|m>,  |vacuum>.

renormalizer/utils/basis.py
liwt31: Indeed. This could be fixed in future commits.

jjren:
In MolList2 multi_e case, I would like the users to take care of the local basis of the electronic site themselves. In some cases, the zero-exciton state is necessary as in linear spectroscopy problems. In some other cases as in the transport problems, only single-exciton states are involved, so the zero-exciton state could be discarded. If someone could like to study the non-linear spectroscopy problems, the two-exciton manifold should also be included. In the multi_e framework, all these problems can be studied flexibly.

The current code is mainly designed for the simple electron case and some parts also support scheme4-like multi-e case. General multi_e case still needs us to redesign some parts like max_entangle, expand_mps, etc.

jjren:
> Apparently, the reason why the routine is not working is that multi-e basis has no creation/annihilation operator because the state with no electron is not represented.
> 
> I wonder why this is the case given that scheme 4 has already demonstrated the zero-electron state is necessary (#27).

In the multi_e case, creation/annihilation operator a^\dagger, a is not well defined.
Only operator like |\Psi_i >< \Psi_j|  or a^\dagger_i a_j is well defined.

liwt31:
> In some other cases as in the transport problems, only single-exciton states are involved, so the zero-exciton state could be discarded. 

At first I thought so, as designed previously in scheme 4. However it turns out that even in transport problems it's not possible to avoid zero-exciton state completely, such as when dealing with initial states. Several workarounds are available, but they would introduce lots of exceptions to the main logic of the code which are quite painful to deal with everyday. For example, if the initial state is obtained by exact thermal propagation, scheme 123 and scheme 4 would require different logic if zero-exciton states are not included in scheme 4.

> If someone could like to study the non-linear spectroscopy problems, the two-exciton manifold should also be included.

I believe introducing zero-exciton states is not contradictory to introducing 2-exciton states.

Overall, introducing zero-exciton states would sometimes increase computational time-cost to a nearly negligible extent, but the gain in programming-efficiency and user-friendliness is massive.

liwt31:
In this commit, I continue to add some doc and clean up code. Most of the work focuses on:
- clean up the test parameter in `test_thermal_prop` of `test_mpdm.py`
- add doc and refactor code for `symbolic_mpo` in `mpo.py` (hope my understanding was correct)

I've basically run through the code and it's really excellent work! My major concerns for the time being are:
-  TDH hybrid algorithm is currently not supported for the new `MolList2`. I wonder if this is temporary or the feature is about to be removed.
- It is clear the code with `MolList2` is much more systematic and comprehensive than `MolList`. It might be a good idea to replace `MolList` with `MolList2`. The original api for `MolList` can be retained, but internally it's converted to (already implemented) and handled as `MolList2`. I think the transition should allow us to delete 2k lines of code.
  However, one of the major drawbacks of the current `MolList2` code is that calculating reduced density matrices is not implemented. There are tricky points around the topic and I think maybe it should be addressed in another PR.
- The last one is the aforementioned zero-exciton states in multi-e basis, which is also not necessary to be dealt with in this PR.

IMO the PR is about to be ready to be merged. No big commits from me anymore.

jjren:
> * TDH hybrid algorithm is currently not supported for the new `MolList2`. I wonder if this is temporary or the feature is about to be removed.

With `MolList2`, we could easily move some unimportant modes to the end of the mps chain, and then tdvp could deal with TD-DMRG/TDH within the same framework by setting some M to just 1 as I've said before.
The only case I could imagine now is to combine TD-DMRG with classical dynamics such as Ehrenfest, which I have made some tests within the same framework with the old ephMPS code.
If someday someone is interested in this direction, keeping it may be useful. But I'm afraid none of us would touch this hybrid quantum/classical things in the foreseeing future.

> * It is clear the code with `MolList2` is much more systematic and comprehensive than `MolList`. It might be a good idea to replace `MolList` with `MolList2`. The original api for `MolList` can be retained, but internally it's converted to (already implemented) and handled as `MolList2`. I think the transition should allow us to delete 2k lines of code.

Yes. But we should do more tests on `MolList2`. Another thing is that deleting 2k lines of the code is not an easy task and will spend some time.

>   However, one of the major drawbacks of the current `MolList2` code is that calculating reduced density matrices is not implemented. There are tricky points around the topic and I think maybe it should be addressed in another PR.

It is done now. I will push it recently.

> * The last one is the aforementioned zero-exciton states in multi-e basis, which is also not necessary to be dealt with in this PR.

We could discuss this issue after the Thursday group meeting.

example/h2o_qc.py
```diff
@@ -0,0 +1,248 @@
...
+
+start = time.time()
+dump_dir = "./"
+job_name = "qc"  #########
+log.register_file_output(dump_dir+job_name+".log", mode="w")
+logger = logging.getLogger(__name__)
+
+spatial_norbs = 7
+spin_norbs = spatial_norbs * 2
+h1e, h2e, nuc = read_fcidump("h2o_fcidump.out", spatial_norbs)
```
liwt31, line 213: This file is not attached.

jjren: Sorry, I forgot it.

jjren:
I think it is time to merge this PR. @liwt31

liwt31:
> I think it is time to merge this PR. @liwt31

I agree. The next step should be a complete migration from `MolList` to `MolList2`. I'll carry out some tests related to my project this month, and then probably we can begin the process in May.

jjren:
> > I think it is time to merge this PR. @liwt31
> 
> I agree. The next step should be a complete migration from `MolList` to `MolList2`. I'll carry out some tests related to my project this month, and then probably we can begin the process in May.

I'm afraid if we completely replace the current MPO construction method with this new algorithm, it would be more time-consuming when constructing the MPOs we have implemented now, especially for the large systems.
We can do some tests on the old and new algorithms.   If they almost cost the same amount of time, I totally agree to remove the old algorithm. But if not (I believe the old algorithm is more efficient), I prefer to keep the old algorithm for cases when MPOs are constructed more frequently in one calculation.

jiangtong1000:
> > > I think it is time to merge this PR. @liwt31
> > 
> > 
> > I agree. The next step should be a complete migration from `MolList` to `MolList2`. I'll carry out some tests related to my project this month, and then probably we can begin the process in May.
> 
> I'm afraid if we completely replace the current MPO construction method with this new algorithm, it would be more time-consuming when constructing the MPOs we have implemented now, especially for the large systems.
> We can do some tests on the old and new algorithms. If they almost cost the same amount of time, I totally agree to remove the old algorithm. But if not (I believe the old algorithm is more efficient), I prefer to keep the old algorithm for cases when MPOs are constructed more frequently in one calculation.

Agree. Just a suggestion, why not keep the frequently-used manually-designed mpo construction and use this general one for other complicated extensions?

liwt31:
> it would be more time-consuming when constructing the MPOs we have implemented now, especially for the large systems.

True. And this is also one of my major concerns. But on my side, I don't see many scenarios that require frequent MPO construction. Besides, I think there're rooms to improve efficiency when MPOs are constructed frequently in one calculation and that is to "remember" the patterns of the interaction to skip the repeated analysis. And yes we should make decision based on benchmark results. 

>  why not keep the frequently-used manually-designed mpo construction and use this general one for other complicated extensions?

In my opinion, removing unnecessary code is very important for the long-term development of any software. The parallel usage of MolList and MolList2 will confuse new users and make code maintenance more difficult. For example, now there're two sets of codes for the calculation of reduced density matrix which share lots of similarities, but their behaviors are subtly different. Trying to modify the code in future could be very painful.
And just to make it clear, the original interface of MolList should be retained (preferably in a more unified way). It's only the internal mechanism that should be changed. 

Even if in the end benchmark results indicates that it's necessary to keep the original code, I would say we should make efforts to "coordinate" the two mechanisms. At least we should invent a better name for MolList2.

jjren:
> Besides, I think there're rooms to improve efficiency when MPOs are constructed frequently in one calculation and that is to "remember" the patterns of the interaction to skip the repeated analysis.

Very good point. I totally agree with you to coordinate these two interfaces.

liwt31:
We can move to issues for future discussion.

PR #84, opened 2020-04-15
enhance expetations calculation

liwt31:
The original code for the calculation of lots of expectations `Mps.expectations`  only supports `mpos` with bond dimension 1. And the MPOs must be only one site different from another standard MPO. This is designed to work with local operators.

Now the code is extended to work with nonlocal operators. The idea is to cache up to N most frequently used intermediate result, where N is the total number of sites. In this way the memory usage is controlled to be approximately the size of the MPS. 

~~I believe the modification has made the `intern` mechanism for `Matrix` no longer meaningful. I'm planning to delete the code and it'll probably show up in the next commit.~~ The statement is incorrect as the current implementation still relies on the fact that equal matrices in MPO share the same id.


liwt31:
The `intern` mechanism for `Matrix` is still removed because it is extremely error-prone

PR #91, opened 2020-05-18
Wyh dev

yhwang627:
1.add 'nac' properties to phonon
2.allow oniste function to create  non-adiabatic coupling operator
3.fix bug in 1&2

### DIFF
--- a/renormalizer/mps/mpo.py
+++ b/renormalizer/mps/mpo.py
@@ -1090,7 +1090,7 @@ class Mpo(MatrixProduct):
         return mpo
 
     @classmethod
-    def onsite(cls, mol_list: MolList, opera, dipole=False, mol_idx_set=None):
+    def onsite(cls, mol_list: MolList, opera, dipole=False, mol_idx_set=None,nrd=False):
         
         if isinstance(mol_list, MolList2):
             # the onsite method is tricky for multi electron case in general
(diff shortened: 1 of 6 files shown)


renormalizer/spectra/finitet.py
```diff
@@ -36,7 +36,9 @@ def __init__(
         gs_shift=0,
         dump_dir: str=None,
         job_name=None,
+        nonradiativedecay=None
```
liwt31, line 39: The default value should be `False` to indicate that this is a Tur/False option.

renormalizer/mps/mpo.py
```diff
@@ -1116,6 +1116,78 @@ def onsite(cls, mol_list: MolList, opera, dipole=False, mol_idx_set=None):
...
+
+                # quantum number part
+                # len(MPO)-1 = len(MPOQN)-2, the L-most site is R-qn
+                mpo.qnidx = len(mpo) - 1
+
+                totnqboson = 0
+                for ph in mol_list[-1].dmrg_phs:
+                    totnqboson += ph.nqboson
+
+                if opera == "a":
```
liwt31, line 1172: Is this condition necessary? `assert opera in ["a"]` in line 1121

liwt31:
Thanks for the effort. Two suggestions:
- It's an important feature and I think corresponding tests are necessary. I noticed there are already some infrastructures in `renormalizer/tests/parameter.py`.
- The onsite MPO code should be much easier to implement with `MolList2` and in the near future `MolList2` will replace `MolList`. Please consider adding another implementation with `MolList2`

renormalizer/mps/mpo.py
jjren: Yes. For nonradiative decay, the transition operator is a de-excitation operator.
But I don't think it is a good choice to put the nrd transition operator in this function `onsite`, because `onsite` now is for electron operator only.
It is better to write a new function (maybe called nrd_transition_operator) or just use `MolList2` method.

liwt31:
Closing this PR because it is better to construct the NAC operator using the MPO generator.

Issue #89, opened 2020-05-16
Another basis for multi electron on one site


liwt31:
The basis `BasisMultiElectron` features flexibility, and a down side is that lots of default behavior are not well-defined. For example, the `gs` method for `Mps` have to be careful with the specific parameter of the basis:
https://github.com/shuaigroup/Renormalizer/blob/ede2b6964fde1ede2eede62880d9938aae4be602/renormalizer/mps/mps.py#L337-L341
This continues the discussions in #82 .

I propose to add another basis for the case where multi electron is on the same site. The quantum number of the basis is constantly `[0] + [1] * (n-1)` and thus creation/annihilation operators and the ground state are well defined. 
The basis should be very useful at least in charge transport problems, which justifies the necessity to add a specific class.
The new basis could be named as `BasisMultiElectronVac` to indicate that the basis includes the vacuum state. 
The transform from scheme 4 to `MolList2` should use this basis by default for maximum backward compatibility. A flag to determine which basis to use might be added if proved necessary.

jiangtong1000:
I thought the zero exciton state can be naturally added to the current MultiElectronBasis, if needed,  by just letting the corresponding sigmaqn = 0, and this is a flexible class.  I am not so sure about what you plan to do.

liwt31:
> I thought the zero exciton state can be naturally added to the current MultiElectronBasis, if needed, by just letting the corresponding sigmaqn = 0, and this is a flexible class. I am not so sure about what you plan to do.

The problem is that the class is **too flexible** at least for charge transport problems. As you can see from the sample code:
https://github.com/shuaigroup/Renormalizer/blob/ede2b6964fde1ede2eede62880d9938aae4be602/renormalizer/mps/mps.py#L337-L341
the ground state of the basis is not well defined because of the flexibility, and a warning should be issued when trying to occupy the ground state. Note that the warning is in a loop so N warnings will popup where N is the total number of e dof. The warning is not the desired behavior in a large number of cases including charge transport problems.

Another example when calculating reduced density matrix:
https://github.com/shuaigroup/Renormalizer/blob/ede2b6964fde1ede2eede62880d9938aae4be602/renormalizer/mps/mps.py#L1538-L1541
Under the flexible basis the reduced density matrix includes the zero exciton state, because the code is not certain whether this information is useful. However in charge transport problems the information is definitely not useful and the default behavior should exclude the zero exciton state.

The worst problem happens when it comes to the onsite operators for `Mpo`:
[...]

PR #93, opened 2020-06-01
Another basis set for multi electron on one site

liwt31:
Solves #89 .
Some points worth noting:
-  `Holstein_model_scheme123` and `Holstein_model_scheme4` in `ModelTranslator` are removed because in `MolList2` these schemes are firstly transformed into general model (or vibronic model) for future MPO construction. 
- The new basis is designed to work with multiple multi-e bases in the same model. The feature has already been put into use for a ladder-shaped Peierls Hamiltonian. This is why `dof_idx_map` is required for the basis.
- The special logic for reduced density matrix calculation when multiple e dofs are on the same site is simplified owing to the new expectation calculation routine #84 .

### DIFF
--- a/renormalizer/model/mlist.py
+++ b/renormalizer/model/mlist.py
@@ -336,11 +336,7 @@ class MolList2:
         
         dipole = {}
         for imol, mol in enumerate(mol_list):
-            if mol_list.scheme < 4:
-                dipole[(f"e_{imol}", )] = mol.dipole
-            elif mol_list.scheme == 4:
-                dipole[(f"e_{imol}", "e_0")] = mol.dipole
-                dipole[("e_0", f"e_{imol}")] = mol.dipole
+            dipole[(f"e_{imol}", )] = mol.dipole
 
         mol_list2 = cls(order, basis, model, model_translator, dipole=dipole)
         mol_list2.map = mapping
--- a/renormalizer/model/mlist.py
+++ b/renormalizer/model/mlist.py
@@ -18,10 +18,6 @@ class ModelTranslator(Enum):
     """
     Available Translator from user input model to renormalizer's internal formats 
     """
-    # from MolList scheme1/2/3
-    Holstein_model_scheme123 = "Holstein Model Scheme123 (MolList)"
-    # from MolList scheme4
-    Holstein_model_scheme4 = "Holstein Model Scheme4 (MolList)"
     # from MolList sbm
     sbm = "Spin Boson Model (MolList)"
     # from MolList2 or MolList augmented with MolList2 parameters
@@ -99,12 +95,24 @@ class MolList2:
                 self.order[o] = i
 
         if isinstance(basis, list):
-            self.basis = basis
+            self.basis: List[ba.BasisSet] = basis
         else:
             assert isinstance(basis, dict)
-            self.basis = [None] * len(self.order)
+            order_value_set = set(self.order.values())
+            if min(order_value_set) != 0 or max(order_value_set) != len(order_value_set) - 1:
+                raise ValueError("order is not continuous integers from 0")
+            self.basis = [None] * (max(self.order.values()) + 1)
             for dof_name, dof_idx in self.order.items():
                 self.basis[dof_idx] = basis[dof_name]
+
+        for b in self.basis:
+            if b.multi_dof:
+                self.multi_dof_basis = True
+                break
+        else:
+            self.multi_dof_basis = False
+
+
         self.model = model
         self.model_translator = model_translator
         # array(n_e, n_e)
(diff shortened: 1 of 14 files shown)


renormalizer/model/mlist.py
```diff
@@ -351,10 +337,10 @@ def e_idx(idx):
         dipole = {}
         for imol, mol in enumerate(mol_list):
             if mol_list.scheme < 4:
-                dipole[(f"e_{e_idx(imol)}", )] = mol.dipole
+                dipole[(f"e_{imol}", )] = mol.dipole
             elif mol_list.scheme == 4:
-                dipole[(f"e_{e_idx(imol)}", "e_0")] = mol.dipole
-                dipole[("e_0", f"e_{e_idx(imol)}")] = mol.dipole
+                dipole[(f"e_{imol}", "e_0")] = mol.dipole
```
jjren, line 342: It is not correct here. The dipole matrix is between vacuum and e_{imol}. 
I think we can only keep `dipole[(f"e_{imol}", )] = mol.dipole` now and come back again when we deal with this kind of "transition" problem someday.

renormalizer/mps/mps.py
```diff
@@ -1495,50 +1495,26 @@ def full_wfn(self) -> np.array:
...
-                l = ones((1, 1))
-                for i in range(e_idx):
-                    mat = self[i].reshape(self[i].shape[0], -1, self[i].shape[-1])
-                    l = tensordot(l, mat, axes=(0, 0))
-                    l = tensordot(l, mat.conj(), axes=([0, 1], [0, 1]))
-                r = ones((1, 1))
-                for i in range(len(self)-1, e_idx, -1):
-                    mat = self[i].reshape(self[i].shape[0], -1, self[i].shape[-1])
-                    r = tensordot(mat, r, axes=(-1, 0))
-                    r = tensordot(r, mat.conj(), axes=([1, 2], [1, 2]))
```
jjren, line 1521: If this part is deleted,  the `BasisMultiElectron` class can not calculate the reduced density matrix anymore.

renormalizer/utils/basis.py
```diff
@@ -150,8 +155,10 @@ def op_mat(self, op):
         op_symbol = op_symbol.split(" ")
         
         if len(op_symbol) == 1:
-            assert op_symbol[0] == "I"
-            mat = np.eye(self.nbas)
+            if op_symbol[0] == "I":
+                mat = np.eye(self.nbas)
+            elif op_symbol[0] == "a" or op_symbol[0] == r"a^\dagger":
+                raise ValueError(f"op_symbol:{op_symbol} is not supported. Try use BasisMultiElectronVac.")
```
jjren, line 161: add `else` is better.

renormalizer/utils/basis.py
```diff
@@ -176,6 +179,75 @@ def op_mat(self, op):
...
+            op_symbol = op_symbol[0]
+            if op_symbol == "I":
+                mat = np.eye(self.nbas)
+            else:
+                mat = np.zeros((self.nbas, self.nbas))
+                op_symbol_term, op_symbol_idx = self._split_sym_idx(op_symbol)
+                if op_symbol_term == r"a^\dagger":
+                    mat[op_symbol_idx, 0] = 1.
+                elif op_symbol_term == r"a":
+                    mat[0, op_symbol_idx] = 1.
```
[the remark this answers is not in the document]
liwt31: This is done in line 246. It is ensured a valid operator is returned.

jjren:
A good idea to add `dof_idx_map` to allow more than one multi-e sites in a DMRG chain.
I believe it would be helpful to reduce the MPO bond dimension of a ladder Hamiltonian.
It is better to add a test for this feature.

renormalizer/utils/basis.py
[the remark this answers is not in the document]
jjren: I do not think so. If an unsupported operator is met, `mat` will be `np.zeros((self.nbas, self.nbas))`. Right?

liwt31:
I've fixed the bug and added a test as required. 
> If this part is deleted, the BasisMultiElectron class can not calculate the reduced density matrix anymore.

I'm not sure I understand what this means exactly, but according to the newly added test it's not a problem to calculate reduced density matrix for BasisMultiElectron.

jjren:
> I've fixed the bug and added a test as required.
> 
> > If this part is deleted, the BasisMultiElectron class can not calculate the reduced density matrix anymore.
> 
> I'm not sure I understand what this means exactly, but according to the newly added test it's not a problem to calculate reduced density matrix for BasisMultiElectron.

I'm wrong. Forget about it.

PR #94, opened 2020-06-09
maintenance work on cv and mpo

jjren:
1. Remove the multiprocessing part inside class `spectra_cv`  to make the `batch_run` independent with this class.
2. Clean up cv subroutines and make it support `MolList2`, see example in `test/test_H_chain.py` which calculates the green function of ab initio 4 atom H-chain.
3. Add a demo of `opt_einsum` in `zerot.py`. The performance is ~10% faster than the hand-designed tensor-contraction strategy.
4. The prescreening of the prefactor in `general_mpo` is changed. Former, it is `np.allclose(prefactor, 0)`, the new is `prefactor == 0`. The change is to avoid neglecting small terms without notice.
5. Optimize the memory usage in `general_mpo`.
6. Add the SHO/dvr basis in `basis.py`. The SHO could support any power of `x` and `p` now.
7. Add a direct solver with `scipy.linalg.eigh` in `mps/solver.py` when the local eigen problem is small.


jjren:
I think cv subroutines still have a room for clean-up and optimization. For example,
1. In `finitet.py`, the operator H^2 is explicitly constructed. It is not efficient apparently.
2. The rule for naming the variables in `zerot.py` and `finitet.py` is better to be the same.
I think @jiangtong1000 could take over this job.

jiangtong1000:
> I think cv subroutines still have a room for clean-up and optimization. For example,
> 
> 1. In `finitet.py`, the operator H^2 is explicitly constructed. It is not efficient apparently.
> 2. The rule for naming the variables in `zerot.py` and `finitet.py` is better to be the same.
>    I think @jiangtong1000 could take over this job when he has time.

OK~

renormalizer/utils/basis.py
```diff
@@ -37,11 +42,23 @@ class BasisSHO(BasisSet):
...
     
-    def __init__(self, omega, nbas, x0=0.):
+    def __init__(self, omega, nbas, x0=0., dvr=False):
         self.omega = omega
         self.x0 = x0  # origin = x0
         super().__init__(nbas, [0] * nbas)
         
+        # calculate x, p power generally, but not efficient for x/p or x^2/p^2
+        # which have been hard-coded already.
+        self.general_xp_power = False
```
liwt31, line 52: Why not add an option for this? This way seems to be error-prone as demonstrated in the bug fix commit 25decac.
Also, the document is not updated.
I can probably help on these things though.

liwt31:
This is a rather big commit and contains several exciting features! I'm reading through the code and preparing a commit to fix several trivial things (typo in docs, comments, etc). I can probably fix my comments myself if they are correct.

renormalizer/utils/basis.py
jjren: Yes. You are right. I will fix this and update the document.

jjren:
> This is a rather big commit and contains several exciting features! I'm reading through the code and preparing a commit to fix several trivial things (typo in docs, comments, etc). I can probably fix my comments myself if they are correct.

Then, I will wait for your commit first.

renormalizer/vibronic/tests/test_pyr4.py
```diff
@@ -214,6 +212,11 @@ def test_pyr_4mode(multi_e, translator):
     mpo = Mpo(mol_list2)
     logger.info(f"mpo_bond_dims:{mpo.bond_dims}")
     mps = Mps.hartree_product_state(mol_list2, condition={"e_1": 1})
+    if dvr:
+        for dof in mol_list2.v_dofs:
+            idx = order[dof]
+            mps[idx] = np.einsum("abc, bd -> adc", mps[idx], basis[idx].dvr_v)
```
liwt31, line 218: These lines of code remind of the idea that maybe we should put the information on what "ground state" is defined in the`Basis` class.
In this example, `BasisSHO` have different representations for "ground state" regarding whether DVR is enabled or not.
And in `hartree_product_state`, for dofs that are not specified in `condition`, the default value is set to the "ground state" provided by the basis class.

jjren: Actually, when DVR is considered, there is a rotation between the basis SHO and SHO_DVR. 
If you only would like to get a hartree_product_state, I think it is easier to just modify `hartree_product_state`. If you'd like to modify BasisSHO,  not only "ground state" but also "1st/2nd/3rd state" should be implemented, because `condition` could be `{"v_1":1}`

liwt31: Makes sense to me. In `hartree_product_state` it should be allowed to specify the state (a vector) instead of just a number.

jjren: Agree.

liwt31:
I've run through half the files and cf65e2d is a quick update. An enhancement on `hartree_product_state` will be included in future commits.

jjren:
> I've run through half the files and [cf65e2d](https://github.com/shuaigroup/Renormalizer/commit/cf65e2de0e9c799dc49291a02ab887f0f78c5fe5) is a quick update. An enhancement on `hartree_product_state` will be included in future commits.

Thanks.

liwt31:
I've made an enhancement on `hartree_product_state` to make it support vector input. To make this change compatible with multi-dof-basis I have to changes on the meaning of `condition` (see the doc). I think it's understandable because hartree_product_state is sensitive to the basis chosen anyway...

jiangtong1000:
- The change of CV part has been reviewed
- finite temperature code is optimized
- add pytest for emission spectrum

renormalizer/cv/spectra_cv.py
```diff
@@ -9,130 +9,181 @@
...
-    '''
-    def run(self):
-        if importlib.util.find_spec("cupy"):
-            multiprocessing.set_start_method('forkserver', force=True)
-        start = time.time()
-        pool = Pool(processes=self.cores)
-        logger.info(f"{len(self.freq_reg)} total frequency to do")
-        logger.info(f"{self.cores} multiprocess parallelization activated")
-        freq_reg = self.freq_reg
-        spectra = []
```
jiangtong1000, line 60: Just to mention. Previously, I did so because I thought that we have to copy many objects of the class "SpectraCv" to make sure that each "object" exist independently and the variables (for example, self.cv_mps) will not be shared by different multiprocesses at the same time, which brings conflicts. However, as you implemented here @jjren, we will not meet such circumstance.  Although I am pretty sure that I did meet troubles at that time by using the logic as yours and I am not able to reproduce the problems any more. Thanks for the improvement.

renormalizer/mps/mps.py
```diff
@@ -199,38 +207,24 @@ def hartree_product_state(cls, mol_list, condition):
...
-            qn_single[isite] = mol_list.basis[isite].sigmaqn[local_state]
-        
-        if not mol_list.multi_dof_basis:
-            for e_dof in mol_list.e_dofs:
-                isite = mol_list.order[e_dof]
-                pdim = mol_list.basis[isite].nbas
-                ms = np.zeros((1, pdim, 1))
-                local_state = condition.get(e_dof,0)
+            if isinstance(local_state, int):
                 ms[0, local_state, 0] = 1.
```
jjren, line 216: It seems that the Hartree product state of  `BasisMultiElectronVac` should be taken with great care because a `gs` is added. Maybe we should add some notes in the doc.

renormalizer/cv/zerot.py
```diff
@@ -27,103 +28,117 @@ class SpectraZtCV(SpectraCv):
...
     def oper_prepare(self, omega):
-        h_mpo = copy.deepcopy(self.mpo)
-        for ibra in range(self.mpo.pbond_list[0]):
-            h_mpo[0][0, ibra, ibra, 0] -= (self.lowest_e + omega)
-        self.a_oper = h_mpo
-
+        # set up a_oper = (H_0 - e0 - omega)
+        identity = Mpo.identity(self.mol_list).scale(-self.e0-omega)
+        self.a_oper = self.h_mpo.add(identity)
+
```
jiangtong1000: This change is intended to avoid doing `deepcopy` of `self.mpo`. But it also brings problems. Firstly, It will increase the bond dim thus the complexity. Secondly, how to make sure the `qntot` of `self.h_mpo` and `identity` the same so that they can add? I noticed that qntot is set to be `0` for `Mpo.identity`.

jjren: Add an MPO with M=1 is not a big problem. The key is that `h_mpo[0][0, ibra, ibra, 0] -= (self.lowest_e + omega)` is not correct in MolList2 because the the first element in MPO[0] is not `H_0` anymore as the MolList1 implementation.
MPO addition does not need that the two operators have s name quantum number.
I think there is no problem that the `qntot` of identity operator is 0.

jiangtong1000: it will check the equality of qn.
```
def add(self, other: "MatrixProduct"):
        assert self.qntot == other.qntot
        assert self.site_num == other.site_num
        assert self.qnidx == other.qnidx
```

jjren: Sorry. I'm wrong. The `qntot` should be the same in addition but not necessary in multiplication.

renormalizer/cv/spectra_cv.py
jjren: To me, the `multiprocessing` is actually a little bit tricky. I'm really not a big fan of it.
This a demo to show that the instance variable would be changed in a child process if `chunksize >1`.  However, If `chunksize =1`, it's ok (imap has default `chunksize=1`).  The same problem has been proposed in https://stackoverflow.com/questions/48332391/instance-methods-with-multiprocessing-pool  
But there is no good answer. It seems that each chunk will have an independent instance because of `pickle`. 
The code works now. But I think maybe we should replace it with `mpi4py` someday because MPI is more sophisticated from my point of view.
```
from multiprocessing import Pool
class c:
    def __init__(self):
        self.a = []

    def print_a(self,i):
        self.a.append(i)
        print(self.a,i, self)
        return i
pool = Pool(processes=2)
hello = c()
for j in pool.imap(hello.print_a, list(range(10)),chunksize=2):
    print(j)
pool.close()
pool.join()
```
output
[0] 0 <__main__.c object at 0x2b0c5bed0d30>
[2] 2 <__main__.c object at 0x2b0c5bed0d30>
[0, 1] 1 <__main__.c object at 0x2b0c5bed0d30>
[2, 3] 3 <__main__.c object at 0x2b0c5bed0d30>
[4] 4 <__main__.c object at 0x2b0c5a76f588>
[6] 6 <__main__.c object at 0x2b0c5a76f588>
[6, 7] 7 <__main__.c object at 0x2b0c5a76f588>
[4, 5] 5 <__main__.c object at 0x2b0c5a76f588>
[8] 8 <__main__.c object at 0x2b0c5a76f748>
[8, 9] 9 <__main__.c object at 0x2b0c5a76f748>

if chunksize=1
[0] 0 <__main__.c object at 0x2b0c5bed0588>
[1] 1 <__main__.c object at 0x2b0c5bed0588>
[2] 2 <__main__.c object at 0x2b0c5a76f748>
[3] 3 <__main__.c object at 0x2b0c5adfe6d8>
[4] 4 <__main__.c object at 0x2b0c5a76f748>
[5] 5 <__main__.c object at 0x2b0c5adfe6d8>
[6] 6 <__main__.c object at 0x2b0c5a76f470>
[7] 7 <__main__.c object at 0x2b0c5a76f588>
[8] 8 <__main__.c object at 0x2b0c5a4d4f28>
[9] 9 <__main__.c object at 0x2b0c5a76f588>

jjren: What we've implemented (use instance method in `map` or `imap`) is actually not a standard way to use multiprocessing in its document (not yet documented at least ).
I guess mpi4py could easily handle it.

jiangtong1000: understood.

jjren:
> * The change of CV part has been reviewed
> * finite temperature code is optimized
> * add pytest for emission spectrum

Hi Tong, why don't you try `opt_einsum` in `finitet.py` since the contraction pattern is much more complicated than zerot ? 
I expect the code may be more readable and the performance may also be better.

jiangtong1000:
> > * The change of CV part has been reviewed
> > * finite temperature code is optimized
> > * add pytest for emission spectrum
> 
> Hi Tong, why don't you try `opt_einsum` in `finitet.py` since the contraction pattern is much more complicated than zerot ?
> I expect the code may be more readable and the performance may also be better.

I haven't tested the performance yet, soon I will make the commit.

liwt31:
4e24227 is  a simple clarification in the `hartree_product_state` doc. 

I'm not particularly familiar with the cv part of code, so I quickly glanced through the commits and found nothing to update. This PR looks good to me now.

jjren:
> [4e24227](https://github.com/shuaigroup/Renormalizer/commit/4e24227fe6053f805342238ca4a4352409f8f58d) is a simple clarification in the `hartree_product_state` doc.
> 
> I'm not particularly familiar with the cv part of code, so I quickly glanced through the commits and found nothing to update. This PR looks good to me now.

@jiangtong1000  Tong, Do you have any comment?

renormalizer/cv/zerot.py
jiangtong1000: should a line `identity.qn_tot = self.h_mpo.qn_tot` be added here?

jjren: > should a line `identity.qn_tot = self.h_mpo.qn_tot` be added here?

I think qntot of Hamiltonian and Identity operator should always be zero.
Thus, I think it is ok here.

jiangtong1000: I am OK with rest of codes.

PR #96, opened 2020-06-24
A quick bugfix in krylov matrix exponential

jjren:
The original implementation has a bug when the size of krylov space is equal to the full space.
This bug will show up when the size of the full space is small.
For example dim=1, alpha is wrong to be [0] after merging  PR #94 where I initialize `j=0`.
dim=2,  the size of the tridiagonal matrix is wrong to be 1.

It is fixed in this PR.

### DIFF
--- a/renormalizer/lib/krylov/krylov.py
+++ b/renormalizer/lib/krylov/krylov.py
@@ -13,7 +13,7 @@ logger = logging.getLogger(__name__)
 
 
 def _expm_krylov(alpha, beta, V, v_norm, dt):
-    # diagonalize Hessenberg matrix
+    # diagonalize Hessenberg matrix (tridiagonal matrix for hermitian matrix A)
     try:
         w_hess, u_hess = eigh_tridiagonal(alpha, beta)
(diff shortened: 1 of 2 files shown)


jjren:
This bug is found by @jiangtong1000 when propagating an uncorrelated single molecular Holstein model.

jiangtong1000:
> This bug is found by @jiangtong1000 when propagating an uncorrelated single molecular Holstein model with `tdvp_ps`.

Actually I just met the problem, just as the case for ground state optimization of small systems, thank you for digging into the root.

jjren:
> > This bug is found by @jiangtong1000 when propagating an uncorrelated single molecular Holstein model with `tdvp_ps`.
> 
> Actually I just met the problem, just as the case for ground state optimization of small systems, thank you for digging into the root.

The problem of ground state optimization of small systems has been fixed in PR #94 in which a direct solver with `scipy.linalg.eigh` has been added.

liwt31:
The bug is malicious and difficult to find out. Thank you for your investigation.

PR #97, opened 2020-06-25
Enhance green-kubo transport and refactor transport module

liwt31:
- `TransportAutoCorr` is renamed to `TransportKubo` and `ChargeTransport` is renamed to `ChargeDiffusionDynamics` for accuracy.
- In `TransportKubo` the current operator are now automatically constructed from `MolList2`.
- If Peielrs terms are present in the model `TransportKubo` will split the current operator into two parts and obtain 5 sets of autocorrelation functions in total (see the doc for details).
- Also completed lots of doc.


renormalizer/transport/kubo.py
```diff
@@ -0,0 +1,346 @@
...
+                    # total term idx should be 0 + 1 + 2 = 3
+                    phonon_term_idx = 3 - term_idx1 - term_idx2
+                    assert term[phonon_term_idx].symbol in (r"b^\dagger + b", "x")
+                symbol1, symbol2 = term[term_idx1].symbol, term[term_idx2].symbol
+                if not {symbol1, symbol2} == {r"a^\dagger", "a"}:
+                    raise ValueError(f"Unknwon symbol: {symbol1}, {symbol2}")
+                if symbol1 == r"a^\dagger":
+                    factor = self.distance_matrix[dof_idx1][dof_idx2]
+                else:
+                    factor = self.distance_matrix[dof_idx2][dof_idx1]
```
jjren, line 190: The model of BasisMultiElectron will result in an error when constructing the current operator with different input formats.
For example 
model[("e_0", )] =["a^\dagger_0 a_1", factor] 
model[("e_0", "e_1")] =["a^\dagger_0 a_1", factor] 
model[("e_0", "e_1")] =["a^\dagger", "a", factor] 
will give the same Hamiltonian, while only the last format will get a correct current operator.
I think it is ok now, because the `transport` is currently not mature. At least we should a `TODO` and `doc` here to make it clear.

renormalizer/transport/tests/test_kubo.py
```diff
@@ -0,0 +1,152 @@
...
+    temperature = Quantity(50000, 'K')
+    compress_config = CompressConfig(CompressCriteria.fixed, max_bonddim=24)
+    evolve_config = EvolveConfig(EvolveMethod.tdvp_ps, adaptive=True, guess_dt=0.5, adaptive_rtol=1e-3)
+    ievolve_config = EvolveConfig(EvolveMethod.tdvp_ps, adaptive=True, guess_dt=-0.1j)
+    kubo = TransportKubo(mol_list, temperature, compress_config=compress_config, ievolve_config=ievolve_config, evolve_config=evolve_config)
+    kubo.evolve(nsteps=5, evolve_time=5)
+    qutip_res = get_qutip_holstein_kubo(mol_list1, temperature, kubo.evolve_times_array)
+    rtol = 5e-2
+    # direct comparison may fail because of different sign
+    assert np.allclose(kubo.auto_corr, qutip_res, rtol=rtol) or np.allclose(kubo.auto_corr, -qutip_res, rtol=rtol)
```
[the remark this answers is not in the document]
liwt31: Well, this is embarrassing. At first I thought it's an implementation detail of `qutip`, but it turned out to be a bug in `BraketPair`. Going to be fixed in the next commit.

renormalizer/transport/kubo.py
```diff
@@ -0,0 +1,346 @@
...
+        else:
+            self.compress_config = compress_config
+
+        self.properties = properties
+        self._auto_corr = []
+        self._auto_corr_deomposition = []
+        super().__init__(evolve_config=evolve_config, dump_dir=dump_dir,
+                job_name=job_name)
+
+    def _construct_current_operator(self):
```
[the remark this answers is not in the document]
liwt31: I've prepared `vibronic_to_general` in `mlist.py` but forgot to use it. Going to be fixed in the next commit.

renormalizer/transport/kubo.py
liwt31: This is a good point, and I think the origin is that with the complexity of `MultiElectron`, the `key:value` structure of `model` is ill-defined. If the problem is not addressed, it will become difficult for future code to deal with the ambiguity.
When merging `MolList` and `MolList2`, I think it's a good chance to improve the interface of `MolList2`. Let's just make it compulsory to set the index inside the operator (`"a^\dagger_0"`) and use a list instead of a dict to represent the model, and `order` can be constructed from `basis` in which each every `BasisSet` knows the external dofs it represents.

----
For this PR I'll fix check the input to preclude the first example. As for the second example, a `ValueError` will be raised, which I think is well enough.

jjren: Maybe we could add a function called `standardize_model` to make the inner format standardized.

liwt31: Yes. The two approaches can be combined. I think users will very much prefer `model=[["a^\dagger_0", "a_1", factor]]` than `model[("e_0", "e_1")] =["a^\dagger_0", "a_1", factor]` , and imagine constructing MPO by `Mpo(mol_list, [["a^\dagger_10", 1]])`. The inner standard format checking could do some sorting over list items, detect duplicate terms, etc.

jjren: Ok.

renormalizer/transport/kubo.py
```diff
@@ -233,21 +247,22 @@ def init_mps(self):
             return BraKetPair(bra_mpdm, ket_mpdm, self.j_oper), BraKetPair(bra_mpdm, ket_mpdm2, self.j_oper2)
 
     def process_mps(self, mps):
+        # add the negative sign because `self.j_oper` is taken to be real
```
jjren, line 250: The -1 here is due to -1.0j * -1.0j of the two j_operators, right?
If so, why don't keep the -1.0j factor to the j_operator since during the time propagation, the mps is complex anyway.
The -1 here looks strange.

liwt31: Yes. j_operator is taken to be real for better performance. In my practice, it is sometimes found that J|ket(0)> is very time and memory consuming. Because |ket(0)> is real, a real j_operator could save half memory and computational cost.

jjren: I see.

Issue #83, opened 2020-02-26
Variational Compress Method


jjren:
Rewrite the variational compress method in the old ephMPS in Renormalizer.
MPO \times MPS with a large bond dimension consumes a lot of memory. Variational compress method could help.

liwt31:
I'm looking forward to this method because in charge transport calculation J @ ket is quite time & memory consuming.
The variational method should help with both accuracy and efficiency.

And a suggestion about the initial guess: SVD compress MPO \times MPS without QR canonicalisation. The compression would be tricky to implement because we would like to avoid explicitly construct the whole MPO \times MPS.

jjren:
I will commit it this week.

jjren:
> And a suggestion about the initial guess: SVD compress MPO \times MPS without QR canonicalisation. The compression would be tricky to implement because we would like to avoid explicitly construct the whole MPO \times MPS.

I tried a similar algorithm (Ann. Phys. 2011, 326, 96−192.) in the old ephMPS code, in which the initial guess is  `canonicalised mpo @ canonicalised mps` and then compressed without full canonicalizsation.
I remembered it does not improve much compared to a random guess but make the code messy. 
Thus in #98 , the initial guess is constructed from compressed_mpo @ compressed_mps and each of them has `M=5`.

liwt31:
Solved by #98

PR #98, opened 2020-07-05
Variational Compression algorithm and some maintenance work

jjren:
The PR contains the following changes:
1. The variational compression algorithm is implemented in `mp.py`, which supports variational compression of `mpo @ mps/mpo/mpdm`. This algorithm could partly solve the memory-bottleneck problem when dealing with `mpo @ mps` and then compression as stated in issue #83 . The related class method `Mpo.contract` is also modified.
   1) The initial guess is constructed from `compressed_mpo @ compressed_mps`  and each of them has `M=5`.
   2) I recommend 2site optimization algorithm if the cost is affordable. 
   3) The related control parameters are set in `CompressConfig` with a `v` prefix.
   4) I add a test in `test_mp.py`.

2. The `solver.py` is refactored and renamed to `gs.py`. Duplicate codes are removed such as `mp._get_big_qn` and `svd_qn.construct_qnmat` (The latter is removed). Add a class method `mp._update_mps`.

3. update some documentation.


doc/source/index.rst
```diff
@@ -5,8 +5,15 @@
 
 Welcome to Renormalizer's documentation!
 ========================================
-Renormalizer is a python package based on tensor network states for electron-phonon
-quantum dynamics.
+Renormalizer is a quantum dynamics package based on density matrix
+renormalization group (DMRG) and its time dependent formulation written in
+Python. All (TD-)DMRG algorithms are implemented within the framework of matrix
+product states (MPS) and matrix product operators (MPO). The package aims to
+provide a light-weight and efficient platform for quantum dynamics simulation
```
liwt31, line 12: The package is not really light-weight 😂 . I'd prefer "a powerful and efficient platform". After all, it's our "aim".

doc/source/ct.rst
```diff
@@ -2,12 +2,12 @@ Charge Transport
 *************************************
 
 Obtain mobility by Green-Kubo formula
-===============================
+=====================================
 .. automodule:: renormalizer.transport.kubo
    :members:
 
 Charge diffusion simulation dynamics
```
liwt31, line 9: The title should be changed to `Charge diffusion dynamics simulation`. It's my fault...

doc/source/index.rst
jjren: Compared to other giant packages, such as heidelberg mctdh, Renormalizer is light-weight. "a powerful and efficient platform" is also Ok.

renormalizer/mps/gs.py
```diff
@@ -104,8 +103,28 @@ def optimize_mps(mps: Mps, mpo: Mpo):
...
+    mpo : renormalizer.mps.Mpo 
+        mpo of Hamiltonian
+    
+    Returns
+    -------
+    energy : list
+        list of energy of each marco sweep.
+        :math:`[e_0, e_0, \cdots, e_0]` if `nroots=1`.
+        :math:`[[e_0, \cdots, e_n], \dots, [e_0, \cdots, e_n]]` if `nroots=n`.
+    mps : renormalizer.mps.Mps
```
liwt31, line 121: Is it necessary to return the mps? The API gives me the impression that the modification is not in-place. If this is for development convenience, the behavior(in-place modification) should be documented.

doc/source/index.rst
liwt31: As far as I know (from Xinxian) the Heidelberg mctdh is big due to its bad architecture. A known issue is that more than 6 layers are not supported because the logics are hardcoded. 
The Itensor package contains about 80k lines of C++ code and Renormalizer contains 10k Python code. We can almost say the "weight" of the package is almost the same with Itensor considering the expressiveness of Python and the extremely low code redundancy in Renormalizer (thanks to constant refactoring).

renormalizer/mps/mpo.py
```diff
@@ -2007,18 +2017,49 @@ def apply(self, mp: MatrixProduct, canonicalise: bool=False) -> MatrixProduct:
...
-                    raise EmptyMatrixError
-        new_mps.canonicalise()
-        new_mps.compress()
+        if algo == "svd":
+            # mapply->canonicalise->compress
+            new_mps = self.apply(mps)
+            if check_emtpy:
+                for mt in new_mps:
+                    if mt.nearly_zero():
+                        raise EmptyMatrixError
```
liwt31, line 2051: If I remember correctly, the `EmptyMatrixError` is the legacy of the days when `scheme4` has no zero exciton state. I'll prepare a PR to remove this feature after this PR is merged.

renormalizer/mps/gs.py
jjren: I think an optimized mps should be returned although the input mps is updated inplace like in `mp.py`, many methods return `self`. I will add some docs.

renormalizer/mps/mpo.py
jjren: Ok

doc/source/index.rst
jjren: Ok, I will modify the description.

renormalizer/mps/tests/test_mp.py
```diff
@@ -59,3 +59,80 @@ def test_environ():
         e = complex(tensordot(l, r, axes=((0, 1, 2), (0, 1, 2)))).real
         assert pytest.approx(e) == mps.expectation(mpo)
 
+
+@pytest.mark.parametrize("comp", (True, False))
+@pytest.mark.parametrize("mp", ("mps", "mpdm", "mpo" ))
+def test_svd_compress(comp, mp):
```
liwt31, line 65: With this test suite, `test_svdqn.py` becomes unnecessary. We can delete `test_svdqn.py` and `test_svd_qn.npz`.

renormalizer/utils/configs.py
```diff
@@ -54,6 +53,10 @@ def __init__(
         threshold: float = 1e-3,
         bonddim_distri: BondDimDistri = BondDimDistri.uniform,
         max_bonddim: int = 32,
+        vmethod: str = "2site",
```
liwt31, line 56: The doc string is not updated

renormalizer/mps/tests/test_mp.py
```diff
@@ -59,3 +59,80 @@ def test_environ():
...
+    mps.compress_config.vprocedure = [[M,1.0],[M,0.2],[M,0.1]]+[[M,0],]*10
+    mps.compress_config.vmethod = "2site"
+    var_mps = mps.variational_compress(mpo, guess=None)
+    dis = var_mps.distance(std_mps)/std_mps.dmrg_norm
+    print(f"var2_mps: {var_mps}, dis: {dis}")
+    assert np.allclose(dis, 0.0, atol=1e-4)
+    assert np.allclose(var_mps.dmrg_norm, std_mps.dmrg_norm, atol=1e-4)
+    
+    # 1site algorithm with 2site result as a guess
+    # 1site algorithm is easy to be trapped in a local minimum
```
liwt31, line 131: Is this an experience or documented in some literature?

renormalizer/mps/tests/test_mp.py
jjren: Agree.

renormalizer/utils/configs.py
jjren: I will update it.

renormalizer/mps/mp.py
```diff
@@ -428,17 +470,323 @@ def compress(self, temp_m_trunc=None, ret_s=False):
...
+                                ltensor, cmo[0], cmo[1], rtensor, cms[0], cms[1],
+                                backend=oe_backend)
+                    elif cms[0].ndim == 4:
+                        # S-a       l-S
+                        #     d   g
+                        # O-b-O-f-O-j-O
+                        #     e   h
+                        # S-c   m   k-S
+                        #     n   p
+                        cout = oe.contract("abc, bdef, fghj, ljk, cenm, mhpk -> adngpl",
```
liwt31, line 620: In the one site branch (line 570), `multi_tensor_contract` is used, while here `oe.contract` is used. I wonder why this is the case because it'll be nice if we can use `oe.contract` without concerns in the future.

liwt31:
@jjren Thanks! I like the idea of `compressed_mpo @ compressed_mps` and the cleaner `gs.py` code. There's not much to modify in the code - primarily docstrings. I'll merge the PR when it's ready.

renormalizer/mps/tests/test_mp.py
jjren: For the electronic system, the standard one-site algorithm usually doesn't work (see Phys. Rev. B 72, 180403(R)). The problem is widely known. 
For Holstein model, in my experience, most of the time one-site algorithm is ok but it is still not that robust compared to the two-site algorithm. In this test, you can try one-site algorithm, the result is not as accurate as two-site algorithm with the same `M`. That is to say, the mps is stuck into a local minima. 
However, since the initial state is too random in this test, I don't know if the performance will be better in a real example.
By the way, for a single mps, I guess the 2site algorithm is not much expensive than 1site algorithm in your problem. If not, I recommend to use 2site algorithm for the initial several sweeps and then 1site algorithm to converge the mps. This strategy is also used in Block (developed by Chan).

liwt31: The information is quite valuable and I think they should be put in the doc string of the method or `CompressConfig`.

renormalizer/mps/mp.py
jjren: In the 1site algorithm, the "optimal" contraction path is known and it is easy to write with `multi_tensor_contract` (`oe.contract` may have an overhead to find a contraction path).
In the 2site algorithm, the contraction path is more complicated and the "optimal" one is not clear (I'm lazy to compare all of them by hand.), so I use `oe.contract` (in my experience when cleaning up `cv`, `oe` performs better in complicated tensor contraction because the overhead to find an "optimal" path is minor.). 
But I haven't done any benchmark on the performance.

renormalizer/mps/tests/test_mp.py
jjren: Agree.

renormalizer/mps/mp.py
liwt31: So, as a rule of thumb, the best practice is to write simple contractions by hand while leaving complicated contractions to `oe.contract`. This makes sense to me. We can follow the guide and update it according to any benchmark results in the future.

jjren: Agree.

jjren:
I change the default `role` to `any` according to https://www.sphinx-doc.org/en/master/usage/restructuredtext/roles.html.
In this way, the cross-reference would be easy to just use \`Mps\` instead of :class:\`~renormalizer.mps.Mps\`.

liwt31:
> So it is better to make html to check whether the document is good or not from now on.

I constantly use this command to build document locally. Also you can use:
```
python -m http.server 
```
in the `html` directory to build a local HTTP server, and then access the doc via `http://localhost:8000`.

PR #104, opened 2020-09-10
DMRG-TDA algorithm

jjren:
1. DMRG-TDA algorithm for excited states.
2. optimize ground state calculation in 
-  better initial guess for state-averaged (SA) calculation
-  the returned mps is the one with the lowest energy. It is more accurate than the mps with active site in the boundary of the chain.
- optimize eigenstate of (\hat{H}-\omega)^2 which could target state near \omega.

3. add another solver (PRIMME https://github.com/primme/primme) for the eigenstate problem. In my current experience, it is better than the pure pyscf.davidson for a difficult problem (for example the density of states near the target is very high).


renormalizer/model/basis.py
```diff
@@ -192,6 +192,8 @@ def op_mat(self, op: Union[Op, str]):
                                          )
             else:
                 mat = np.diag(self.dvr_x**2)
+        elif op_symbol == "x x":
```
liwt31, line 195: Cases like this can be simplified by:
```
elif op_symbol == "x^2" or op_symbol=="x x":
```

renormalizer/mps/lib.py
```diff
@@ -97,6 +116,54 @@ def read(self, domain: str, siteidx: int):
         return asxp(self._virtual_disk[(domain, siteidx)])
 
 
+def contract_one_site_multi_mpo(environ, ms, mos, domain, ms_conj=None):
```
liwt31, line 119: The function name and documentation implies contraction over more than 2 MPOs is also supported. 
I don't know if the circumstance will appear in the future, but I think it is possible to write a generic routine based on `opt_einsum`. Another benefit is that `contract_one_site_multi_mpo` and `contract_one_site` can be united.

This is not urgent and can be postponed to future works.

renormalizer/model/basis.py
jjren: It's because another flag self.general_xp_power would lead to inconvenience.

renormalizer/mps/lib.py
jjren: In `test_mp.py`, I add a test `test_environ_multi_mpo` to make sure `contract_one_site_multi_mpo` could generalize `contract_one_site`.  As you said `cv` could use it to construct the environment. In this commit, I use this function to optimize the ground state of (H-\omega)^2. 
I can do this maintenance in the future.

liwt31: > In this commit, I use this function to optimize the ground state of (H-\omega)^2.

Speaking of which, is it feasible to construct `(H-\omega)^2` as an MPO? *Maybe* the resulting scaling is lower.

renormalizer/mps/mp.py
```diff
@@ -166,16 +166,22 @@ def is_right_canon(self):
         return self.qnidx == 0
 
     def ensure_left_canon(self, rtol=1e-5, atol=1e-8):
-        if not self.check_left_canonical(rtol, atol):
+        if (not self.check_left_canonical(rtol, atol)) or self.to_right or\
```
liwt31, line 169: There is a minor optimization at this point: the conditions that do not need extensive calculation such as `self.to_right` could be moved before those that require extensive calculation such as `self.check_left_canonical`

renormalizer/mps/mps.py
```diff
@@ -1240,6 +1240,156 @@ def full_wfn(self) -> np.array:
...
+                    elif self[kms].ndim == 4:
+                        tensor = tensordot(tensor, self[kms], ([2,3,4],[0,1,2]))
+                
+                rtensor = R_component[jms]
+                res = tensordot(tensor, rtensor,
+                        ([2,3],[0,1])).transpose(0,2,1,3)
+                rdm[(ims, jms)] = asnumpy(res.reshape(res.shape[0]*res.shape[1],-1))
+        return rdm
+    
+    def calc_orbital_entropy(self, entropy_type):
```
liwt31, line 1337: The interface and documentation become confusing considering the old `calc_vn_entropy` code.
How about `calc_entropy(self, entropy_type)` where `entropy_type` can be `"1o"`, `"2o"` or `"bond"`?

renormalizer/mps/mps.py
```diff
@@ -1240,6 +1240,156 @@ def full_wfn(self) -> np.array:
...
+            else:
+                tensor = tensordot(tensor, ms, ([0,-1,-2],[0,-1,-2]))
+            assert xp.allclose(tensor, tensor.T.conj())
+            rdm[ims] = asnumpy(tensor)
+
+        return rdm
+    
+    def calc_2ordm(self):
+        r""" Calcuate 2-orbital reduced density matrix
+        :math: `\rho_{ij} = \textrm{Tr}_{k \neq i, k \neq j} | \Psi \rangle \langle \Psi |`
```
liwt31, line 1283: This equation fails to render in the doc.

renormalizer/mps/mps.py
```diff
@@ -1240,6 +1240,156 @@ def full_wfn(self) -> np.array:
...
+
+        entropy = {}
+        for key, dm in rdm.items():
+            w, v = scipy.linalg.eigh(dm)
+            w[w<=0] = 1e-100
+            entro = np.sum(-w*np.log(w))
+            entropy[key] = entro
+        return entropy
+    
+    def calc_orbital_mutual_entropy(self):
```
liwt31, line 1371: I think a test case is missing for the method. `Mpo.identity(self.mol_list)` appears in the code but `mol_list` is no longer an attribute for `mp` now.

renormalizer/mps/lib.py
jjren: You mean we manually expand (H-omega)^2 as a single operator (eliminate and combine some terms are possible) or just multiply two MPO (H-\omega)? If the latter, the scaling is higher (think about constructing the environment) and meanwhile the memory usage is also higher.

renormalizer/mps/mp.py
jjren: good point.

renormalizer/mps/mps.py
jjren: In quantum dynamics, people may prefer ``DoF``. In QC, it is obscure to use ``DoF`` to denote electrons. The most precise definition here is ``site``. Do you think we should replace it with ``site``?

renormalizer/mps/mps.py
jjren: I will check it.

renormalizer/mps/mps.py
jjren: I will add one.

renormalizer/mps/mps.py
liwt31: I like `"1siste"` and `"2site"`. 1-site entropy and 2-site entropy are pretty self-explanatory and they look DMRGish.

renormalizer/mps/mps.py
liwt31: The problem is probably because of the additional space in `:math: `\rho_{ij}`

renormalizer/mps/mps.py
jjren: Ok.

renormalizer/mps/lib.py
liwt31: I mean the first one, taking advantage of the general MPO construction routine and `Op` multiplication operations. Maybe we can let an undergraduate student test it :)

jjren: It is possible that the scaling would be smaller in some cases (maybe in the Holstein-like model). Then, it could be used to reduce the cost of the `cv` algorithm. 
However, in QC (or vibrational Hamiltonian, or other general long-range Hamiltonians too), the H^2 has MPO with O(N^4) just as the direct multiplication of two H (O(N^2) \times O(N^2)).  In these cases, the scaling is worse.

renormalizer/mps/tda.py
```diff
@@ -0,0 +1,505 @@
...
+            
+    
+    def analysis_dominant_config(self, thresh=0.8, alias=None, tda_m_trunc=20,
+            return_compressed_mps=False):
+        r""" analyze the dominant configuration of each tda root.
+            The algorithm is to compress the tda wavefunction to a rank-1 Hatree
+            state and get the ci coefficient of the largest configuaration.
+            Then, the configuration is substracted from the tda wavefunction and
+            redo the first step to get the second largest configuration. The
+            two steps continue until the thresh is achieved.
```
liwt31, line 380: Several typos: "Hatree", "configuaration", "substracted"

liwt31:
I'm not familiar with the DMRG-TDA algorithm but they seem just fine to me. Maybe you'll get the chance to introduce the idea to us in the future.

jjren (commented):
Sorry for the delayed update.

renormalizer/mps/mps.py
```diff
@@ -274,14 +290,22 @@ def ground_state(cls, model: Model, max_entangled: bool):
                         nzeros = basis.sigmaqn.count(0)
                         qn = basis.sigmaqn[dof.split("_")[1]]
                         if qn == 0:
-                            ms[0, dof.split("_")[1], 0] = 1. / np.sqrt(nzeros)
+                            if normalize:
+                                # Todo: here the dof_name only supports XXX_0, XXX_1
```
liwt31, line 294: Maybe a quick fix is to use the index of `0` elements in `basis.sigmaqn`?

liwt31:
Seems fine to me. Thanks, Jiajun 👍

liwt31 (approved):

renormalizer/mps/mps.py
jjren: I will submit a quick pull-request soon.

PR #109, opened 2020-10-19
fix mps.load bug

liwt31:
After TDH hybrid algorithm is removed (#61 ), the `coeff` for `mps` is changed from type `list` to type `complex` or `float`.
However, when an `Mps` (or `MpDm`) is loaded from an `npz` file, the datatype for `coeff` is cast to a NumPy array of size 1 * 1. This is the default behavior for the `npz` archive. 
This bug could lead to severe problems because the code logic is based on the assumption that `coeff` is a number instead of an array.  For example, in `Mps.metacopy`  it is written:
```
new.coeff = self.coeff
```
and then `new` and `self` would share the same `coeff` array.

A quick fix is to use the [`tolist` function of NumPy](https://numpy.org/doc/stable/reference/generated/numpy.ndarray.tolist.html)

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 Renormalizer is a python package based on tensor network states for electron-phonon quantum dynamics.
 
-[![Build Status](https://travis-ci.org/shuaigroup/Renormalizer.svg?branch=master)](https://travis-ci.org/shuaigroup/Renormalizer)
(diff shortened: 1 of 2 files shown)


jjren:
I cannot figure out why `tolist` could solve this problem.   `new` and `self` would still share the same list.

jjren:
I suggest to use numpy.ndarray.item.
https://numpy.org/doc/stable/reference/generated/numpy.ndarray.item.html

liwt31:
Fixed the outdated badge in README.md by the way

jjren (approved):

PR #110, opened 2020-11-30
Add one-particle spectral function for charge transport problems

liwt31:
- A new class `SpectralFunctionZT` for the one-particle spectral function at zero temperature. The finite temperature effect can be taken into account using the thermal field dynamics algorithm mentioned in the document and exemplified in the test case.
- The algorithm is only valid for a translational invariant one-dimensional system. To restrict its application and simplify model definition for the very common case, a new `Model` for translational invariant one dimensional system `TI1DModel` is added.
- The `todo` for removing `h_mpo` argument in `ThermalProp`  is based on the fact that the `h_mpo` for every step of time evolution is actually constructed from the model of the `h_mpo`:
https://github.com/shuaigroup/Renormalizer/blob/9aeaad9190b11c55898fbbe51c025b9a92471044/renormalizer/mps/thermalprop.py#L100-L102
However the constructed `h_mpo` is not necessarily the same as the `h_mpo` in the argument. Since in `Mpo.__init__` we actually allow construct MPO apart from the Hamiltonian. I've got a plan to improve the API and I think I'm going to open an issue for it as it involves a drastic change of a commonly used API.

### DIFF
--- a/renormalizer/model/model.py
+++ b/renormalizer/model/model.py
@@ -29,13 +29,19 @@ class Model:
         Contains the transition dipole matrix element. The key is the dof name.
     """
     def __init__(self, basis: List[BasisSet], ham_terms: List[Op], dipole: Dict = None):
-
+        all_dof_list = []
+        for local_basis in basis:
+                all_dof_list.extend(local_basis.dofs)
+        if len(all_dof_list) != len(set(all_dof_list)):
+            raise ValueError("Duplicate DoF definition found in the basis list.")
         self.basis: List[BasisSet] = basis
         # alias
         self.dof_to_siteidx = self.order = {}
+        self.dof_to_basis = {}
         for siteidx, ba in enumerate(basis):
             for dof_name in ba.dofs:
                 self.dof_to_siteidx[dof_name] = siteidx
+                self.dof_to_basis[dof_name] = ba
 
         self.ham_terms: List[Op] = self.check_operator_terms(ham_terms)
         # array(n_e, n_e)
@@ -216,7 +222,7 @@ class HolsteinModel(Model):
             \rm{ph}_{n/2+1, 0}, \rm{ph}_{n/2+1, 1}, \cdots]
 
     periodic : bool
-        Whether use periodical boundary condition when constructing `j_matrix`
+        Whether use periodical boundary condition when constructing ``j_matrix``
         from :class:`~renormalizer.utils.Quantity`. Default is ``False``.
     """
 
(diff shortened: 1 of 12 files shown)


renormalizer/transport/spectral_function.py
```diff
@@ -0,0 +1,133 @@
...
+        latest_ket_mpdm = prev_ket_mpdm.evolve(self.h_mpo, evolve_dt)
+        return (prev_bra_mpdm, latest_ket_mpdm)
+
+    def get_dump_dict(self):
+        dump_dict = dict()
+        dump_dict['temperature'] = self.temperature.as_au()
+        dump_dict['time series'] = self.evolve_times
+        dump_dict["G array"] = self.G_array
+        ne = self.model.n_edofs
+        ka = np.linspace(0, np.pi, ne // 2 + 1).reshape(1, 1, -1)
```
jjren, line 128: I think ka = np.linspace(-pi,pi,ne+1).reshape(1, 1, -1).
when ne is 2n-1, the result is same as the current definition, when ne is 2n it is different.

renormalizer/model/model.py
```diff
@@ -387,6 +393,75 @@ def __init__(self, epsilon: Quantity, delta: Quantity, ph_list: List[Phonon], di
...
+        Terms of the system local Hamiltonian :math:`\hat h_i` in sum-of-product form.
+        DoF names should be consistent with the ``basis`` argument.
+    nonlocal_ham_terms : :class:`list` of :class:`~renormalizer.model.Op`
+        Terms of system nonlocal Hamiltonian :math:`\hat h_{i, i+j}`.
+        To indicate the IDs of the unit cells that are involved in the nonlocal interaction,
+        the DoF name in ``basis`` should be transformed to a two-element-tuple, in which
+        the first element is an integer :math:`j`, and the second element is the original DoF name.
+        For example, if one unit cell contains one electron DoF with name ``e``,
+        a nearest neighbour hopping interaction
+        should take the form ``Op(r"a^\dagger a", [(0, "e"), (1, "e")])``
```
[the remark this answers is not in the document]
liwt31: Yes, but it'll be a very unusual way to express the terms. Do you think it's necessary to clarify it in the document?

renormalizer/transport/spectral_function.py
liwt31: Actually, it is when ne=2n that the two sets of code are equivalent. I think a better way is to assert ne is even since there's no point looking for the "absent k=0" trouble. Of course, the reason of the striction and the specific choice of the k points should be documented.

renormalizer/model/model.py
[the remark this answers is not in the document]
jjren: Yes, I think it is better to make it clear in the document for the other users.

renormalizer/transport/spectral_function.py
jjren: Yes. I'm wrong. When ne=2n, the results are the same. 
Yes. It is better to assert "ne is even" and add some docs. 
I don't understand the "absent k=0" trouble.

liwt31: In some cases it is desired to analyze spectral function at the k=0 point.

jjren: The k=0 point is always there. but the k=pi point is not guaranteed (yes when ne=2n, no when ne=2n+1).

liwt31: Yes, you're right. I get confused. Still, I think it's better to make ne an even number, since the k=pi point is also quite special.

Besides, it makes me wondering if the odd/even problem is important in transport problems. I always use odd numbers as a convention from the diffusion dynamics ages.

jjren: If ne is approaching the thermodynamic limit, I think it is not important for ne to be odd or even.

renormalizer/model/model.py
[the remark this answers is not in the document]
liwt31: The documentation has been updated.

renormalizer/transport/spectral_function.py
liwt31: In the end, I decided to make it compatible with even and odd `ne` since in the test case its `ne=3` and `ne=4` is too time consuming.

renormalizer/model/model.py
```diff
@@ -399,11 +399,11 @@ class TI1DModel(Model):
...
 
     .. math::
-        \hat H = \sum_i(\hat h_i + \sum_j \hat h_{i, i+j})
+        \hat H = \sum_i(\hat h_i + \sum_j \hat h_{i,j})
 
     where :math:`\hat h_i` is the local Hamiltonian acting on one single unit cell
-    and :math:`\hat h_{i, i+j}` represents the interaction between the :math:`i` th cell
-    and the :math:`i+j` th cell.
+    and :math:`\hat h_{i,j}` represents the :math:`j` th interaction between the :math:`i` th cell
+    and other unit cells.
```
jjren, line 406: The doc here confuses me. 
I think `j` does not represent the `j` th interaction. Instead, it is the index of  `j`th site, representing the two-site interaction between `i`th and `j`th sites.

liwt31: The model is not limited to two-unitcell interaction and `h_{ij}` appears to be the most general way to express it.

jjren (approved):

renormalizer/model/model.py
jjren: Got it!

PR #113, opened 2021-03-19
Update on ThermalProp interface, FT-CT interface and optimize SVD

liwt31:
The PR contains three commits (and two bugfix commits):
- Update the interface for `ThermalProp` as mentioned in #110 . `h_mpo` is replaced by `h_mpo_model`, which is used to construct `h_mpo` at every imaginary time step. 
- Allow manually setting thermal dump path in `TransportKubo`.
- Optimize `Csvd`. When `full_matrices` is set to `True`, the routine is particularly time-consuming for matrices with imbalanced shape, such as (100, 100\*40\*40). However, in almost all cases it is not necessary to construct the complete matrix (100\*40\*40, 100\*40\*40) as required by `full_matrices=True`. The optimization contained in this PR is to set `full_matrices=False` and add additional Gram-Schmidt orthonormalized basis to the matrices after SVD decomposition. The number of the basis is set to be equal to the dimension of the smaller index (100 in the above dimension). The update should allow dynamic growth of bond dimension while gaining tremendous efficiency when the shape of the matrix to be decomposed is imbalanced.
    The same optimization can be carried out for QR but for now QR is only used with `full_matrices=False`, so the code for QR is not modified.

At first I decided to remove the finite temperature interface for `ChargeDiffusionDynamics` since using Bogoliubov transformation is more convenient. However `ChargeDiffusionDynamics` is based on `HolsteinModel` and using Bogoliubov transformation is not straight-forward.

### DIFF
--- a/renormalizer/mps/mp.py
+++ b/renormalizer/mps/mp.py
@@ -463,8 +463,7 @@ class MatrixProduct:
         """
 
         if mpo is None:
-            logger.info("Recommend to use svd to compress a single mps/mpo/mpdm.")
-            raise NotImplementedError
+            raise NotImplementedError("Recommend to use svd to compress a single mps/mpo/mpdm.")
         
         if guess is None:
             # a minimal representation of self and mpo
--- a/renormalizer/mps/mp.py
+++ b/renormalizer/mps/mp.py
@@ -408,7 +408,7 @@ class MatrixProduct:
         for idx in self.iter_idx_list(full=False):
             mt: Matrix = self[idx]
             qnbigl, qnbigr, _ = self._get_big_qn([idx])
-            u, sigma, qnlset, v, sigma, qnrset = svd_qn.Csvd(
+            u, sigma, qnlset, v, sigma, qnrset = svd_qn.svd_qn(
                 mt.array,
                 qnbigl,
                 qnbigr,
@@ -647,7 +647,7 @@ class MatrixProduct:
         if type(cstruct) is not list:
             # SVD method
             # full_matrices = True here to enable increase the bond dimension
-            Uset, SUset, qnlnew, Vset, SVset, qnrnew = svd_qn.Csvd(
+            Uset, SUset, qnlnew, Vset, SVset, qnrnew = svd_qn.svd_qn(
                 asnumpy(cstruct), qnbigl, qnbigr, self.qntot, system=system
             )
 
@@ -685,8 +685,9 @@ class MatrixProduct:
                         axes=(range(qnbigl.ndim), range(qnbigl.ndim)),
                     )
             ddm /= len(cstruct)
-            Uset, Sset, qnnew = svd_qn.Csvd(asnumpy(ddm), qnbigl, qnbigr, self.qntot,
-                    system=system, ddm=True)
+            Uset, Sset, qnnew = svd_qn.eigh_qn(
+                asnumpy(ddm), qnbigl, qnbigr, self.qntot, system=system
+            )
             ms, msdim, msqn, compms = select_basis(
                 Uset, Sset, qnnew, None, Mmax, percent=percent
             )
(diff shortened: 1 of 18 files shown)


renormalizer/mps/mp.py
```diff
@@ -591,16 +589,15 @@ def variational_compress(self, mpo=None, guess=None):
...
+                logger.info(f"Variation compress relative error: {error}")
+                if error < mps.compress_config.vrtol:
+                    logger.info("Variational compress is converged!")
+                    break
             
             mps_old = mps.copy()
-
-        if converged:
-            logger.info("Variational compress is converged!")
         else:
```
jjren, line 601: I can not find the `if` corresponding to the `else` here. Please check it.

renormalizer/mps/svd_qn.py
```diff
@@ -1,10 +1,67 @@
...
+
+def optimized_svd(a, full_matrices):
+    # optimize performance when ``full_matrices=True`` and the shape of ``a`` is extremely unbalanced
+    # The idea is to construct only a limited number of orthogonal basis rather than all of them
+    # (which are not necessary in most cases)
+    m, n = a.shape
+
+    # whether do the optimization
+    # here 1/3 and 3 are only empirical
+    opt = not (1 / 3 < m / n < 3)
```
jjren, line 20: Why do not add a `opt` flag in `optimized_svd` and `Csvd`? 
if opt is None, opt = not (1 / 3 < m / n < 3). 
In this case, if we need real `full_matrices` in the future, we can just use `opt=False`.

renormalizer/mps/mp.py
liwt31: This is a syntax for the `if` clause. The code within the clause is executed if the loop is not stopped by ``break``. Example:
```
# Determine if 7 is a prime number
n = 7
for x in range(2, n):
    if n % x == 0:
        print(n, 'equals', x, '*', n//x)
        break
else:
    # loop fell through without finding a factor
    print(n, ' is a prime number')
```
[Official document](https://docs.python.org/3/tutorial/controlflow.html#break-and-continue-statements-and-else-clauses-on-loops)

renormalizer/mps/svd_qn.py
liwt31: Agreed. The `opt` option also serves as documentation for the `Csvd` function.

renormalizer/mps/svd_qn.py
```diff
@@ -1,10 +1,67 @@
...
+    m, n = u.shape
+    assert 2 * n < m
+    assert np.allclose(u.T.conj() @ u, np.eye(n))
+    res = np.zeros((m, 2 * n), dtype=u.dtype)
+    res[:, :n] = u
+    for i in range(n):
+        a = np.random.rand(m)
+        a = a - res @ (res.T.conj() @ a)
+        a /= np.linalg.norm(a)
+        res[:, n + i] = a
```
jjren, line 60: For higher efficiency with vectorization, I think about another algorithm to construct `res` without the manual Gram-Schmidt orthogonalization.
```
a = np.random.rand(m,n)
a = a - res @ (res.T.conj() @ a)
q, r = scipy.linalg.qr(a, mode='economic') 
res[:, n:] =  q
```

liwt31: Do you mean:
```
a = np.random.rand(m,n)
a = a - u @ (u.T.conj() @ a)
q, r = scipy.linalg.qr(a, mode='economic') 
res[:, n:] =  q
```
I think there should be a measurable speed up, I'll carry out some tests for sure.

liwt31: For large matrices (100000 * 100), old routine costs 1.64 s, while the QR version costs 0.682 s.
For small matrices (100 * 10), old routine costs 420 mus, while the QR version costs 301 mus.

liwt31:
At first I was only trying to add some document. But then I realized It is probably better to split the `ddm` treatment out of the `Csvd` function. I also carried out some refactoring by the way. The code becomes shorter with less nested `if-else` conditions.

renormalizer/mps/svd_qn.py
jjren: That's good.

jjren (approved):

PR #115, opened 2021-04-09
refactor gs.py

liwt31:
This commit involves a lot of lines of code but the actual modification is minor. Just split the `optimize_mps` function to multiple functions in order to hopefully make the logic easier to understand.

### DIFF
--- a/renormalizer/mps/gs.py
+++ b/renormalizer/mps/gs.py
@@ -6,45 +6,25 @@ several lowest excited states with state-averaged algorithm.
 
 """
 
-from typing import Tuple, List
+from typing import Tuple, List, Union
 import logging
-
-from renormalizer.lib import davidson
-from renormalizer.mps.backend import xp, USE_GPU
-from renormalizer.mps.matrix import (
-    multi_tensor_contract,
-    tensordot,
-    asnumpy, 
-    asxp)
-from renormalizer.mps import Mpo, Mps
-from renormalizer.mps.lib import Environ, cvec2cmat
-from renormalizer.utils import Quantity
 
 import numpy as np
 import scipy
 import opt_einsum as oe
 import primme
 
+from renormalizer.lib import davidson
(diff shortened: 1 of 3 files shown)


liwt31:
Fixed a bug regarding `res_mps_idx`. In the original code it is assumed that the the active site idx of the last sweep should be the  active site idx of the next sweep. This is not the case for 2-site method, in which the idx is differed by 2.

renormalizer/mps/gs.py
```diff
@@ -134,365 +96,436 @@ def optimize_mps(mps: Mps, mpo: Mpo, omega :float =None) -> Tuple[List, Mps]:
...
+        logger.debug(f"energy: {e}")
+        micro_iteration_result.append(e)
+
+        cstruct = cvec2cmat(cshape, c, qnmat, mps.qntot, nroots=nroots)
+
+        # store the "optimal" mps (usually in the middle of each sweep)
+        should_store_res_mps: bool = False
+        if last_opt_e_idx is not None:
+            if method == "1site":
+                res_mps_idx = last_opt_e_idx
```
jjren, line 249: I think the `res_mps_idx` is still not correct even for 1-site algorithm. For example: 4 sites,0,1,2,3 sweeping from right to left. The micro_iteration_result is [e3,e2,e1,e0]. if the minimum energy is e1, the `res_mps_idx=2`, however the correct res_mps_idx for the next left-to-right sweep is `imps=1`.

jjren: Maybe replace the current type of `micro_iteration_result` with a dictionary `{(active_site):e,...}` will make the bug easy to solve.

liwt31: I see your point here. The culprit is `micro_iteration_result.append(e)` even though the iteration is not from left to right.

jjren: Yes, you are right.

jjren (approved):

PR #117, opened 2021-06-21
dump/load method in mp.py to allow dump and load Mpo and refine SineDVR basis

jjren:


renormalizer/mps/mp.py
```diff
@@ -934,6 +934,38 @@ def _array2mt(self, array, idx, allow_dump=True):
     
     def build_empty_mp(self, num):
         self._mp = [[None]] * num
+    
+    def dump(self, fname, other_attrs=[]):
```
liwt31, line 938: Not a big fan of using `[]` as the default argument (see [this post](https://www.toptal.com/python/top-10-mistakes-that-python-programmers-make))

renormalizer/model/basis.py
```diff
@@ -382,45 +383,253 @@ def op_mat(self, op: Union[Op, str]):
         if op_symbol == "I":
             mat = np.eye(self.nbas)
         
-        elif op_symbol.split("^")[0] == "x":
+        elif op_symbol.split("^")[0] == "x" and " " not in op_symbol:
```
liwt31, line 386: Is these code just for checking or it does something different from the good old `mat = np.diag(self.dvr_x ** moment)`?

liwt31:
This commit seems to involve a large number of equations. Are they derived or from literature? If it's the latter case, adding the reference to the document could be helping.

jjren:
The matrix element is derived through some manual integrations.

renormalizer/model/basis.py
jjren: The modification ensures analytical accuracy for the operators such as `x p`, `x^2`, `x^3`, `x p^2` etc.

liwt31: Am I to understood that even for `x` `mat = np.diag(self.dvr_x)` is not accurate?

jjren: For `x`, it is exact. But for `x^2`, `x^3`, etc, it is not.

renormalizer/mps/mp.py
jjren: fixed

liwt31 (approved):

PR #118, opened 2021-07-03
update doc and requirements

jjren:


jjren:
In this PR, I want to update the dependent packages such as numpy and scipy to the latest version. But I find that python 3.6 only supports up to numpy 1.19  and scipy 1.5.  
I don't know if it is necessary to update to python 3.7 since numpy 1.19 (even 1.16) and scipy 1.5 work well with Renormalizer.

liwt31:
Great work. Currently I see no need for drop Python 3.6 support. We can postpone this update for a major modification to our dependencies in order to reduce the (number of) inconveniences.

liwt31:
Great work. Currently I see no need for drop Python 3.6 support. We can postpone this update for a major modification to our dependencies in order to reduce the (number of) inconveniences.

PR #119, opened 2021-10-04
improve mpo construction performance

jjren:
improve the performance of the MPO auto construction algorithm.
When I tested spin-boson model with 1000 modes, I found the hotspot is in
```
coord = np.nonzero(symbol_table == op_to_dict(op))
```
which is used to convert the original Op table to a table with integers to represent different operators.
I optimized this part in this PR.
The new version costs 2 min: 42 s for the upper Hamiltonian.
The old version costs 9 min: 09s.


liwt31 (approved):
This happens to be what I've done in my local version :)

PR #121, opened 2022-01-01
Implement on the fly swapping

liwt31:
This PR implements the on the fly swapping (OFS) algorithms for DOF ordering optimization.

The swapping for MPS is implemented in `mp._update_mps`.
The swapping for MPO along with the old symbolic MPO code is moved to a new file `symbolic_mpo`.

There are four possible applications of the OFS:
- static DMRG (implemented)
- TDVP-PS2 (implemented)
- Variational Compress (not implemented)
- CV (not implemented)

The configuration of OFS is included in `CompressConfig`. Two additional options are added to the class.


renormalizer/mps/mp.py
```diff
@@ -645,11 +603,68 @@ def _update_mps(self, cstruct, cidx, qnbigl, qnbigr, Mmax, percent=0):
...
+                qnbigl2, qnbigr2, _ = self._get_big_qn(cidx, swap=True)
+                if cstruct.ndim == 4:
+                    cstruct2 = asnumpy(cstruct).transpose(0, 2, 1, 3)
+                else:
+                    assert cstruct.ndim == 6
+                    cstruct2 = asnumpy(cstruct).transpose(0, 3, 4, 1, 2, 5)
+                if self.compress_config.ofs_swap_jw:
+                    assert cstruct2.ndim == 4
+                    cstruct2 = cstruct2.copy()
+                    cstruct2[:, 1, 1, :] = -cstruct2[:, 1, 1, :]
```
jjren, line 630: 1. This phase is assuming that |0>/|up> is unoccupied, |1>/|down> is occupied. This choice depends on the actual Jordan-Wigner transformation we use. I think it is better to add some assert or at least some comment somewhere.
2.  why is the cstruct2  copied?

renormalizer/mps/mp.py
```diff
@@ -645,11 +603,68 @@ def _update_mps(self, cstruct, cidx, qnbigl, qnbigr, Mmax, percent=0):
...
+                    Uset, SUset, qnlnew, Vset, SVset, qnrnew = \
+                        Uset2, SUset2, qnlnew2, Vset2, SVset2, qnrnew2
+                    qnbigl, qnbigr, cstruct = qnbigl2, qnbigr2, cstruct2
+                    new_basis = self.model.basis.copy()
+                    new_basis[cidx[0]:cidx[1] + 1] = reversed(self.model.basis[cidx[0]:cidx[1] + 1])
+                    # previously cached MPOs are destroyed.
+                    # Not sure what is the best way: swap all cached MPOs or simply reconstruct them
+                    # Need some additional testing at production level calculation
+                    self.model: Model = Model(new_basis, self.model.ham_terms, self.model.dipole, self.model.output_ordering)
+                logger.debug(f"DOF ordering: {[b.dof for b in self.model.basis]}")
```
jjren, line 667: What are previously cached MPOs?

renormalizer/mps/symbolic_mpo.py
```diff
@@ -0,0 +1,645 @@
...
+    out_ops1, out_ops2, out_ops3 = out_ops_list
+
+    # the expanded sum-of-product form of the operator represented by the two site MPO
+    # i.e. the expanded form of out_ops3
+    # the number of items is equal to the index of bond 3
+    # each item [factor, out_ops[1]idx, primary_ops @ site 1, primary_ops @ site 2]
+    out_ops3_expanded: List[List[ExpandedOp]] = []
+    for out_op_sum_list in out_ops3:
+        out_ops3_expanded.append(expand_out_op_sum_list(out_ops2, out_op_sum_list))
+        del out_op_sum_list
```
jjren, line 576: Is the `del` necessary?

renormalizer/mps/symbolic_mpo.py
```diff
@@ -0,0 +1,645 @@
...
+    assert len(new_out_ops3) == len(primary_ops) - n_primary_ops == len(auxiliary_dummy_primary_ops)
+    assert len(new_out_ops[-1]) == 1
+    for dummy_op in new_out_ops[-1][0]:
+        idx1, idx2 = dummy_op.symbol
+        idx2 -= n_primary_ops
+        new_out_ops3[idx2] = new_out_ops3_unsorted[idx1]
+        if dummy_op.factor != 1:
+            for i, op in enumerate(new_out_ops3[idx2]):
+                new_out_ops3[idx2][i] = OpTuple(symbol=op.symbol, qn=op.qn, factor=op.factor * dummy_op.factor)
+        del dummy_op, idx1, idx2
```
jjren, line 624: I do not understand why we should explicitly delete these objects?

renormalizer/mps/symbolic_mpo.py
```diff
@@ -0,0 +1,645 @@
...
+
+    # translate the numbers into symbolic Matrix Operator
+    def compose_symbolic_matrix_operator(in_ops, out_ops):
+        mo = [[[] for o in range(len(out_ops))] for i in range(len(in_ops))]
+        for iop, out_op in enumerate(out_ops):
+            for composed_op in out_op:
+                in_idx = composed_op.symbol[0]
+                op = primary_ops[composed_op.symbol[1]]
+                mo[in_idx][iop].append(composed_op.factor * op)
+        return mo
```
jjren, line 639: This function looks like a duplicate of the one above

jjren (changes requested):
:D

renormalizer/mps/mp.py
liwt31: 1. OK. 
2. because it is possible `cstruct2` shares the same underlying data with `cstruct`. Makes a copy can avoid overwrite `cstruct`

renormalizer/mps/mp.py
liwt31: Operators for the calculation of e and ph occupations and so on.

renormalizer/mps/symbolic_mpo.py
liwt31: When I wrote the code I find that this statement helps me remember `out_op_sum_list` is a local variable. It indeed seems awkward after a while...

jjren: ok. I get your point.

renormalizer/mps/mp.py
liwt31: Yes. When swapping happens very frequently, I think it is best to simply reconstruct the MPOs after one sweep. If the swapping rarely happens, then naturally we can just swap the MPOs. I think we need more experience to decide what is the best practice in the general case.

renormalizer/mps/symbolic_mpo.py
jjren: I guess the reason is the same as above.

jjren (approved):

PR #123, opened 2022-03-20
make dependency on primme optional

liwt31:
New users of renormalizer often struggle to install primme, which is not frequently used. In this PR I make the dependency on `primme` optional. It is removed from `requiements.txt`. But if it is correctly installed, renormalizer still supports the package. The Travis CI test configurations are updated such that two different sets of tests are run. The first is without `primme` (default) and the second is with `primme`. 

Also fix a small bug in OFS.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -7,5 +7,4 @@ PyYAML==5.4.*
 recommonmark==0.7.*
 sphinx_rtd_theme==0.5.*
 opt_einsum==3.3.*
-primme==3.2.*
 qutip==4.6.*
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -51,7 +51,7 @@ jobs:
           command: cd doc; make html
 
       - persist_to_workspace:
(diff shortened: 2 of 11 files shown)


liwt31:
It seems we have run out of credit on Travis CI. So I switch to Circle CI.

requirements.txt
```diff
@@ -7,5 +7,4 @@ PyYAML==5.4.*
 recommonmark==0.7.*
 sphinx_rtd_theme==0.5.*
 opt_einsum==3.3.*
-primme==3.2.*
```
jjren, line 10: It is better to keep the `primme` package here but with a comment  # in order to remind the user to install it if possible.

liwt31: I don't think a comment here will help "remind" any user. It is better added to the installation doc, and the old one is very outdated.

jjren: at least remind us :joy:

jjren (approved):

PR #124, opened 2022-03-24
add support for time-dependent H and VSCF

jjren:
In this PR, there are several updates:
1. support for time-dependent Hamiltonian
2. generic Runge-Kutta solver
3. vibrational SCF with a DMRG sweep algorithm
4. refactorize mps.normalize


renormalizer/mps/mps.py
```diff
@@ -639,31 +644,170 @@ def expand_bond_dimension(self, hint_mpo=None, coef=1e-10, include_ex=True):
...
-        if np.iscomplex(evolve_dt):
-            new_mps.normalize(1.0)
+        if normalize:
+            if np.iscomplex(evolve_dt):
+                new_mps.normalize(1.0)
+            else:
+                new_mps.normalize(None)
+        return new_mps
+    
+    def _evolve_prop_and_compress_tdrk4(self, mpo, evolve_dt) -> "Mps":
```
liwt31, line 667: Why this should be a separate method? What is the difference between this method and `_evolve_prop_and_compress_tdrk` with `rk_solver == "C_RK4"`?

renormalizer/vibration/vscf.py
```diff
@@ -0,0 +1,119 @@
...
+            logger.info("load h_mpo form model.mpos")
+            self.h_mpo = model.mpos["h_mpo"]
+        else:
+            self.h_mpo = Mpo(model)
+        if mps is None:
+            self.mps = Mps.hartree_product_state(self.model, dict())
+        else:
+            self.mps = mps
+    
+    def kernel(self, nsweeps=100):
```
liwt31, line 37: A lot of the code is in duplication with `gs.py`. Is it possible to call `optimize_mps` in this function?

renormalizer/mps/mps.py
jjren: `_evolve_prop_and_compress_tdrk` is more general. 
`_evolve_prop_and_compress_tdrk4` is legecy code.  
Currently, I keep it for test.

renormalizer/vibration/vscf.py
jjren: It is not convenient to call `optimize_mps` unless `optimize_mps` is adapted to vscf,  including the convergence criteria, how to solve the eigenproblem, etc.
Considering vscf is rarely used and the current sweep-like implementation is an ad-hoc method, I think it is better to separate it from the more complex DMRG code.

renormalizer/mps/mps.py
```diff
@@ -564,23 +549,34 @@ def metacopy(self) -> "Mps":
...
+        '''
+
+        if kind == "mps_only":
+            new_coeff = self.coeff
+        elif kind == "mps_and_coeff":
+            new_coeff = self.coeff / np.linalg.norm(self.coeff)
+        elif kind == "mps_norm_to_coeff":
+            new_coeff = self.coeff * self.mp_norm
+        else:
+            raise ValueError(f"kind={kind} is valid.")
```
liwt31, line 574: invalid? 😂

jjren: sorry, it is invalid.

renormalizer/vibration/vscf.py
liwt31: OK. The primary differences are that 1) for the eigenproblem, all roots are solved and stored, but the MPS is only updated by the lowest eigenstate and 2) for the convergence criteria, both energy and coefficient are taken into account. Am I right?
If it is really "ad-hoc" then in the future we should replaced it by another standard implementation. Otherwise maintaining two pieces of similar code would be troublesome.

liwt31:
Looks good to me. The factored normalization is much more clearer and readable than the original version. Would you like to patch the typo, or merge now and fix the typo in the future?

renormalizer/vibration/vscf.py
jjren: For vscf, the standard implementation is to update the coeff of each site together like HF.

jjren: Yes, you are right.

jjren:
I've fixed the typo.

liwt31 (approved):

PR #132, opened 2022-09-17
Change import behavior

liwt31:
- Make Renormalizer easier to use for other projects when imported as one of the many packages.
    - Renormalizer by default will not set the number of threads for NumPy. This is something to remind the group members and write into the document. Accordingly, change the NumPy import error into a warning
    - Change the logger behavior. Avoid using root logger.
- Add better support for 32-bit backend. Make some of the tests that originally can not pass or very time-consuming in 32-bit backend work for 32-bit backend.
- Optimize performance in circle CI by parrallel pytest. Now the whole tests take less than 6 min.

### DIFF
--- a/renormalizer/mps/backend.py
+++ b/renormalizer/mps/backend.py
@@ -67,8 +67,10 @@ class Backend:
         self.first_mp = False
         self._real_dtype = None
         self._complex_dtype = None
-        #self.use_32bits()
-        self.use_64bits()
+        if os.environ.get("RENO_FP32") is None:
+            self.use_64bits()
+        else:
+            self.use_32bits()
 
     def free_all_blocks(self):
         if not USE_GPU:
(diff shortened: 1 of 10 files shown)


renormalizer/mps/backend.py
```diff
@@ -130,5 +132,12 @@ def dtypes(self):
     def dtypes(self, target):
         self.real_dtype, self.complex_dtype = target
 
+    @property
+    def canonical_atol(self):
+        if self.is_32bits:
+            return 1e-4
+        else:
+            return 1e-5
```
jjren, line 140: why `atol` is changed from 1e-8 to 1e-5 for double precision?

jjren: I get it. Only the criterion for the off-diagonal elements is looser.

jjren (approved):

Issue #143, opened 2022-10-21
Change the default time evolution method


liwt31:
Currently, the default time evolution method is `prop_and_compress`, which is not very popular. Maybe it's time to change it to TDVP_PS or TDVP_VMF?

jjren:
RK4 is actually more black box than TDVP_PS and TDVP_VMF to get not bad results (of course not the best) for most Hamiltonians.
Because TDVP methods need a good/correct initialization (renormalized space expansion) to avoid the evolution stuck in a wrong subspace. It depends on if the user understands and runs the algorithm correctly.
For example, in the transport problem, all the one-exciton state should be included in the initial renormalized space carefully.
If not, you will get an artificial localization solution.
I don't know whether 2-site TDVP_PS can avoid this kind of problem. But RK4 is more robust, at least get a "correct" answer.
Do you think so?

liwt31:
Makes sense. However I'd like to add that the step control in TDVP_VMF is a unique advantage.

Then it again boils down to the documentation. We should list all the possible schemes and their (dis)advantages. Some of them are published with well-supported data (like the 2020JCP work). Some of them are more empirical.

jjren:
I totally agree with you about improving the documentation for all the evolution algorithms. 
It is better if you can write a template and set a doc structure, then I can add content.

For the current situation, I think we can switch the default algorithm to P&C TD RK (general explicit RK propagator for both time-dependent/independent H), with which the time-step can be controlled adaptively (e.g Cash-Karp45, RKF45).

liwt31:
> I don't know whether 2-site TDVP_PS can avoid this kind of problem. 

I think not if the electronic sites are separated by phonon sites. Maybe this is worth documenting too.

> It is better if you can write a template and set a doc structure, then I can add content.

No problem

> we can switch the default algorithm to P&C TD RK

OK. Could you do it while adding the documentation? I assume you're more familiar with the setup. Honestly I rarely use this method and I'm under the impression that the step-size control for P&C RKF is not particularly reliable. Sometimes the step size simply keeps shrinking to infinitely small in one step.

jjren:
> OK. Could you do it while adding the documentation? I assume you're more familiar with the setup. Honestly I rarely use this method and I'm under the impression that the step-size control for P&C RKF is not particularly reliable. Sometimes the step size simply keeps shrinking to infinitely small in one step.

[...]

PR #144, opened 2022-10-26
Update document

liwt31:
- Fix the Heisenberg tutorial bug
- Add time evolution page. #143 
- Add the FAQ page

update @ 2023.4.8 @jjren 
- remove the m_max argument in _update_mps in mp.py. 
First, because it is duplicated with mps.compress_config.
The more important reason is that if we want to adaptively control the bond dimension by truncation threshold in the two-site tdvp-ps or ground state algorithm, the former implementation can not achieve that.
(What I haven't modified is OFS algorithm because one loss function depends on m_max, so I added an assertion to make sure the bond dimension is fixed.)

- add scipy ivp solver company to Krylov solver in the tdvp related evolution algorithm.
This will make the evolution algorithm suitable for non-Hermitian operators.

- add quadrature in SineDVR basis by using the sympy expression to support more operators.
This feature is only experimental, not fully tested.

- bump up dependencies (numpy, scipy) in requirement.txt

### DIFF
--- a/renormalizer/cv/spectra_cv.py
+++ b/renormalizer/cv/spectra_cv.py
@@ -5,11 +5,12 @@ import numpy as np
 from multiprocessing import Pool
 import multiprocessing
 from renormalizer.mps import Mpo
-from renormalizer.utils import Quantity
+from renormalizer.utils import Quantity, CompressCriteria, CompressConfig
 from renormalizer.utils.elementop import construct_e_op_dict, ph_op_matrix
 import importlib.util
 import logging
 from typing import List
+
 
 logger = logging.getLogger(__name__)
 
@@ -103,6 +104,8 @@ class SpectraCv(object):
         else:
             self.cv_mps = cv_mps
         
+        self.cv_mps.compress_config = CompressConfig(CompressCriteria.fixed, max_bonddim=m_max)
+
         # results
         self.hop_time = []
         self.macro_iteration_result = []
--- a/renormalizer/model/basis.py
+++ b/renormalizer/model/basis.py
@@ -5,7 +5,8 @@ import scipy.linalg
 import scipy.special
 import itertools
 import logging
-
+import sympy as sp
+import scipy.integrate
 
 logger = logging.getLogger(__name__)
 
@@ -106,6 +107,14 @@ class BasisSet:
         raise NotImplementedError
 
 
+class BasisDummy(BasisSet):
+    """
+    Dummy basis set used in soft truncated TNPI
+    """
+    def  __init__(self, dof, nbas):
+        super().__init__(dof, nbas, [0] * nbas)
+
+
 class BasisSHO(BasisSet):
     """
     simple harmonic oscillator basis set
@@ -131,8 +140,8 @@ class BasisSHO(BasisSet):
 
         self.general_xp_power = general_xp_power
 
-        # whether under recurssion
-        self._recurssion_flag = 0
+        # whether under recursion
+        self._recursion_flag = 0
 
         self.dvr = False
         self.dvr_x = None  # the expectation value of x on SHO_dvr
@@ -148,12 +157,14 @@ class BasisSHO(BasisSet):
         if not isinstance(op, Op):
             op = Op(op, None)
         op_symbol, op_factor = op.symbol, op.factor
+        
+        op_symbol = op_symbol.replace("partialx", "dx")
 
         if op_symbol in ["b", "b b", r"b^\dagger", r"b^\dagger b^\dagger", r"b^\dagger b", r"b b^\dagger", r"b^\dagger+b"]:
-            if self._recurssion_flag == 0 and not np.allclose(self.x0, 0):
+            if self._recursion_flag == 0 and not np.allclose(self.x0, 0):
                 logger.warning("the second quantization doesn't support nonzero x0")
 
-        self._recurssion_flag += 1
+        self._recursion_flag += 1
 
         # so many if-else might be a potential performance problem in the future
         # changing to lazy-evaluation dict should be better
@@ -296,8 +307,8 @@ class BasisSHO(BasisSet):
                     + self.op_mat(r"b b^\dagger")
                     - self.op_mat(r"b^\dagger b"))
 
-        elif op_symbol == "x partialx":
-            # x partialx is real, while x p is imaginary
+        elif op_symbol == "x dx":
+            # x dx is real, while x p is imaginary
             mat = (self.op_mat("x p") / -1.0j).real
 
         elif op_symbol == "p x":
@@ -306,13 +317,13 @@ class BasisSHO(BasisSet):
                     - self.op_mat(r"b b^\dagger")
                     + self.op_mat(r"b^\dagger b"))
 
-        elif op_symbol == "partialx x":
+        elif op_symbol == "dx x":
             mat = (self.op_mat("p x") / -1.0j).real
 
-        elif op_symbol == "partialx":
+        elif op_symbol == "dx":
             mat = (self.op_mat("p") / -1.0j).real
 
-        elif op_symbol in ["partialx^2", "partialx partialx"]:
+        elif op_symbol in ["dx^2", "dx dx"]:
             mat = self.op_mat("p^2") * -1
         elif op_symbol == "I":
             mat = np.eye(self.nbas)
@@ -324,7 +335,7 @@ class BasisSHO(BasisSet):
         else:
             raise ValueError(f"op_symbol:{op_symbol} is not supported. ")
 
-        self._recurssion_flag -= 1
+        self._recursion_flag -= 1
         return mat * op_factor
 
     def copy(self, new_dof):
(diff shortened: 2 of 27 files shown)


liwt31:
I may work on the tutorial module while you finish the time evolution part. @jjren

jjren:
Hi Weitang @liwt31 , at the beginning of this PR, you mentioned the Heisenberg tutorial bug, but I didn't find any changes to the Heisenberg tutorial. What I found is that the jupyter notebook of Heisenberg tutorial is supposed to move into  doc/source/tutorial, but it doesn't.  
Did I do something wrong during the force push ...
or you haven't move the jupyter notebook of Heisenberg into doc/source/tutorial?

jjren:
My another question is what the difference between the three CI tests.
From my understanding, ci / test (pull request) is the PR branch + master, ci / test (push) is the PR branch only. Then what is ci/circleci ?

jjren:
> Hi Weitang @liwt31 , at the beginning of this PR, you mentioned the Heisenberg tutorial bug, but I didn't find any changes to the Heisenberg tutorial. What I found is that the jupyter notebook of Heisenberg tutorial is supposed to move into doc/source/tutorial, but it doesn't. Did I do something wrong during the force push ... or you haven't move the jupyter notebook of Heisenberg into doc/source/tutorial?

I found that in the successful ci/circleci run, it showed that 
```
Running Sphinx v5.3.0
copying ../../example/1D-Heisenberg.ipynb
making output directory... done
```
How did it happen? I can't find a rule anywhere in the package to do this.

jjren:
> > Hi Weitang @liwt31 , at the beginning of this PR, you mentioned the Heisenberg tutorial bug, but I didn't find any changes to the Heisenberg tutorial. What I found is that the jupyter notebook of Heisenberg tutorial is supposed to move into doc/source/tutorial, but it doesn't. Did I do something wrong during the force push ... or you haven't move the jupyter notebook of Heisenberg into doc/source/tutorial?
> 
> I found that in the successful ci/circleci run, it showed that
> 
> ```
> Running Sphinx v5.3.0
> copying ../../example/1D-Heisenberg.ipynb
> making output directory... done
> ```
> 
> How did it happen? I can't find a rule anywhere in the package to do this.

Oh, finally, I found it in `/home/jjren/Code/Renormalizer/doc/source/conf.py` !
The remaining question is that
- the bug in Heisenberg mode tutorial
- the  pandoc seems not installed successfully in the two ci/test.

liwt31:
- CI/CircleCI is based on the PR branch only. We are in the middle of switching from circleci to GitHub actions.
- I can't really remember the bug exactly. Probably related to documentation generation and not the content in the notebook
- CircleCI passes because in this branch the installation of pandoc is added to .circleci/config.yaml. The same lines should be added to .github/workflow/ci.yaml

renormalizer/__init__.py
```diff
@@ -6,7 +6,7 @@
 import warnings
 
 
-reno_num_threads = os.environ.get("RENO_NUM_THREADS")
+reno_num_threads = os.environ.get("RENO_NUM_THREADS", 1)
```
liwt31: The default behavior is changed on purpose in #132 . I strongly oppose setting the default number of threads when importing renormalizer. The reason is that this is a self-centered feature and is bad for the applicability of renormalizer.

When renormalizer is imported as a regular package for another project (such as TenCirChem), setting the number of threads by default can cause unexpected behavior. For example, TenCirChem itself may have set the `"MKL_NUM_THREADS"` to a certain value for its purpose, and an experienced user can also set `"MKL_NUM_THREADS"` directly without referring to the documentation or source code of renormalizer. But the value will be overwritten silently by renormalizer when importing the package and there's not even a good way to prevent this behavior.

Also, the best practice for production-level calculation is to set the number of threads to 1, 2 or 4 based on the details of the calculation. Setting the default number of threads to 1 is more like a development shortcut. In fact, in my development environment, I hardcoded the number of threads to be 1 in `__init__.py` but I never commit this change.

liwt31: This is also for consistency with popular computational software such as pyscf (https://github.com/pyscf/pyscf/issues/540)

renormalizer/cv/spectra_cv.py
```diff
@@ -103,6 +104,8 @@ def __init__(
         else:
             self.cv_mps = cv_mps
         
+        self.cv_mps.compress_config = CompressConfig(CompressCriteria.fixed, max_bonddim=m_max)
```
liwt31, line 107: Is `self.m_max` still used? (resolve directly if yes)

renormalizer/model/basis.py
```diff
@@ -424,58 +436,67 @@ def __init__(self, dof, nbas, xi, xf, endpoint=False):
...
+
+        elif op_symbol == "x^2":
+             mat = self._I()*self.xi**2+self._u()*self.xi*2+self._uu()
+        
+        elif op_symbol == "x^3":
+            mat = self._I()*self.xi**3 + 3*self._uu()*self.xi + 3*self._u()*self.xi**2 + self._uuu()
+        
         elif set(op_symbol.split(" ")) == set("x"):
             moment = len(op_symbol.split(" "))
+            self._recursion_flag = 1
```
liwt31, line 496: Is it better to use `self._recursion_flag += 1` just as BasisSHO for deep recursion?

renormalizer/mps/gs.py
```diff
@@ -107,12 +107,21 @@ def optimize_mps(mps: Mps, mpo: Mpo, omega: float = None) -> Tuple[List, Mps]:
     # determines the index of active site of the returned mps
     opt_e_idx: int = None
     res_mps: Union[Mps, List[Mps]] = None
-    for isweep, (mmax, percent) in enumerate(mps.optimize_config.procedure):
+    for isweep, (compress_config, percent) in enumerate(mps.optimize_config.procedure):
         logger.debug(f"isweep: {isweep}")
-        logger.debug(f"mmax, percent: {mmax}, {percent}")
+        
+        if isinstance(compress_config, CompressConfig):
+            mps.compress_config = compress_config
```
liwt31, line 114: What about the original `mps.compress_config`? Should we record it and then setting it back after the optimization?

renormalizer/mps/mps.py
```diff
@@ -1190,9 +1199,21 @@ def func1(y):
                                     coef, ovlp_inv1=S_L_inv_list[imps+1],
                                     ovlp_inv0=S_L_inv_list[imps], ovlp0=S_L_list[imps])
                             return func(0, y)
+
```
liwt31, line 1202: Is it better to wrap this routine as a function?

liwt31:
In general looks good! Only need to resolve the problem of the default thread number.

renormalizer/mps/gs.py
jjren: We control the bond dimension in gs.py with mmax and the original mps.compress_config is not used. I guess it is the default one. So is it necessary to set it back?

renormalizer/__init__.py
jjren: Got it. I will recover it.

renormalizer/mps/gs.py
liwt31: The user may have set this value for future time evolution. pseudo-code:
```
mps.compress_config = ...
mps.optimize_config = ...
mps = optimize_mps(mps, mpo, procedure)
mps = excitation_mpo @ mps
mps = mps.evolve_with_rk(tau)
```

renormalizer/cv/spectra_cv.py
jjren: Yes. It is used in initialized random state and select basis in finite temperature algorithm.
It is not hard to replace all of them with information from compres_config.
I didn't do it because of the time. I just make sure that all the tests are passed. :cry:

renormalizer/model/basis.py
jjren: ok. I will do it.

renormalizer/mps/mps.py
jjren: Did you mean func1 is not necessary?

renormalizer/mps/gs.py
jjren: I got it.

renormalizer/cv/spectra_cv.py
liwt31: OK then no problem

renormalizer/mps/mps.py
liwt31: I mean L1203 to L1215, which has a lot of duplication in this commit

renormalizer/mps/gs.py
jjren: solved.

renormalizer/__init__.py
jjren: solveed

renormalizer/model/basis.py
jjren: solved

jjren:
Finally. :turtle:

liwt31:
great stuff, thanks!

PR #149, opened 2023-03-13
[WIP] Basis tree tensornetwork algorithm

liwt31:
This is more like a preview than an actual PR. Feedbacks are welcome.
Implemented features:
- Automatic tree tensornetwork operator (TTO) construction
- 2site optimization algorithm

Please see the test cases for how to use them.

Obviously, there are a lot of important features that remain to be implemented. The following ones have a higher priority
- QN support
- VMF (imaginary) time evolution

After that, we may have a proper comparison/benchmark with (ML-)MCTDH

On the longer term I'm going to look at
- GPU support
- PS time evolution
- Finite temperature (Could be *very* tedious)
- TTO/TTS (tree tensornetwork states) manipulations
- Compression
- Documentation

Of course, one may imagine a lot of other features such as excited states optimization algorithm and technical enhancements like dumping to or loading from hard disk. In the foreseeable future, I'm afraid I'll only work on ones that will enable finite temperature mobility calculation due to my limited time.

### DIFF
--- a/renormalizer/__init__.py
+++ b/renormalizer/__init__.py
@@ -32,7 +32,7 @@ del os, log_level, init_log
 
 
 # user interfaces
-from renormalizer.model import Model, HolsteinModel, SpinBosonModel, TI1DModel, Op
+from renormalizer.model import Model, HolsteinModel, SpinBosonModel, TI1DModel, Op, OpSum
 from renormalizer.model.basis import BasisSHO, BasisHopsBoson, BasisSineDVR, BasisMultiElectron, \
     BasisMultiElectronVac, BasisSimpleElectron, BasisHalfSpin
 from renormalizer.mps import Mpo, Mps, optimize_mps
--- a/renormalizer/model/model.py
+++ b/renormalizer/model/model.py
@@ -515,3 +515,13 @@ def load_from_dict(param, scheme, lam: bool):
     j_constant = Quantity(*param["j constant"])
     model = HolsteinModel([Mol(Quantity(0), ph_list)] * param["mol num"], j_constant, scheme)
     return model, temperature
+
+
+def heisenberg_ops(nspin):
+    ham_terms = []
+    for ispin in range(nspin - 1):
+        op1 = Op("sigma_z sigma_z", [ispin, ispin + 1], 1.0 / 4)
+        op2 = Op("sigma_+ sigma_-", [ispin, ispin + 1], 1.0 / 2)
+        op3 = Op("sigma_- sigma_+", [ispin, ispin + 1], 1.0 / 2)
+        ham_terms.extend([op1, op2, op3])
+    return ham_terms
--- a/renormalizer/mps/mps.py
+++ b/renormalizer/mps/mps.py
@@ -193,7 +193,7 @@ class Mps(MatrixProduct):
                     if a basis contains ``"e_1"``, ``"e_2"`` and ``"e_3"``,
                     ``{"e_1": 2}`` (``{"e_1": [0, 0, 1]}``) means ``"e_3"`` is occupied and
                     ``{"e_1": 1}`` (``{"e_1": [0, 1, 0]}``) means ``"e_2"`` is occupied.
-                    Be aware that in :class:`renormalizer.utils.basis.BasisMultiElectronVac` the vacuum state
+                    Be aware that in :class:`renormalizer.BasisMultiElectronVac` the vacuum state
                     is added to the ``0`` index.
 
         Returns:
--- a/renormalizer/mps/symbolic_mpo.py
+++ b/renormalizer/mps/symbolic_mpo.py
@@ -9,6 +9,7 @@ import scipy
 import scipy.sparse
 
 from renormalizer.model import Model
+from renormalizer.model.basis import BasisSet
 from renormalizer.model.op import Op
 from renormalizer.lib import bipartite_vertex_cover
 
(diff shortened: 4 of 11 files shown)


jjren:
excellent PR! The tree tensornetwork operator part deserves a new paper.
I will quickly review the current code and leave some suggestions.

liwt31:
Circle CI is failing again. Will migrate to Github actions completely.

liwt31:
Supports multiple basis sets on the same site. Should be useful for finite-temperature algorithms, mode combinations and so on.

Will stick to the original roadmap after this "digression"

liwt31:
Add QN support for TTO/TTS and random TTS generation. Also, add the Davidson eigensolver and the primme eigensolver (configuration not supported yet).

Somehow the eigenvalue problem in TTN is more difficult to solve. The scipy solver fails to converge in the CI, and the Davidson solver will converge to wrong result on my machine, starting from multiple different initial guesses. The primme solver works pretty well though. The small $M$ could be the problem,  but the scaling over $M$ is not very favorable so higher $M$ is not tested yet.

liwt31:
> Add QN support for TTO/TTS and random TTS generation. Also, add the Davidson eigensolver and the primme eigensolver (configuration not supported yet).
> 
> Somehow the eigenvalue problem in TTN is more difficult to solve. The scipy solver fails to converge in the CI, and the Davidson solver will converge to wrong result on my machine, starting from multiple different initial guesses. The primme solver works pretty well though. The small M could be the problem, but the scaling over M is not very favorable so higher M is not tested yet.

It turns out to be a bug: the environment is not properly canonicalized. Fixed in the latest commit

liwt31:
Add VMF (imaginary) time evolution for tree tensor network. QN for VMF has not been implemented yet.
Matrix Unfolding is also not implemented.
I tested various algorithms for regularization of $S$, and adding $\epsilon$ to the diagonal elements and then calling `scipy.linalg.pinv` seems to be the most robust one. The following ones will quickly result in NaN
- Add $\epsilon \exp{(-S/\epsilon)}$ to $S$ rather than $\epsilon I$
- Invert regularized $S$ by inverting eigenvalues
- Invert regularized $S$ by `scipy.linalg.inv`

Although the methods above only yield very slightly (around 1e-10) different $\tilde{S}^{-1}$ from the implemented approach.
Manually setting the first step size for the RK solver also helps sometimes. Overall the problem is rather complicated.

jjren:
> Add QN support for TTO/TTS and random TTS generation. Also, add the Davidson eigensolver and the primme eigensolver (configuration not supported yet).
> 
> Somehow the eigenvalue problem in TTN is more difficult to solve. The scipy solver fails to converge in the CI, and the Davidson solver will converge to wrong result on my machine, starting from multiple different initial guesses. The primme solver works pretty well though. The small M could be the problem, but the scaling over M is not very favorable so higher M is not tested yet.

This is actually why I use primme when I implemented the TDA algorithm. I found that primme is much more robust.

jjren:
> Add VMF (imaginary) time evolution for tree tensor network. QN for VMF has not been implemented yet. Matrix Unfolding is also not implemented. I tested various algorithms for regularization of S, and adding ϵ to the diagonal elements and then calling `scipy.linalg.pinv` seems to be the most robust one. The following ones will quickly result in NaN
> 
> * Add ϵexp⁡(−S/ϵ) to S rather than ϵI
> * Invert regularized S by inverting eigenvalues
> * Invert regularized S by `scipy.linalg.inv`
> 
> Although the methods above only yield very slightly (around 1e-10) different S~−1 from the implemented approach. Manually setting the first step size for the RK solver also helps sometimes. Overall the problem is rather complicated.

That's weird. Will a larger epsilon solve this problem?
Maybe we should try diffeq.jl/diffeqpy, because it supports more solvers (e.g. stiffness auto detections).

liwt31:
> > Add VMF (imaginary) time evolution for tree tensor network. QN for VMF has not been implemented yet. Matrix Unfolding is also not implemented. I tested various algorithms for regularization of S, and adding ϵ to the diagonal elements and then calling `scipy.linalg.pinv` seems to be the most robust one. The following ones will quickly result in NaN
> > 
> > * Add ϵexp⁡(−S/ϵ) to S rather than ϵI
> > * Invert regularized S by inverting eigenvalues
> > * Invert regularized S by `scipy.linalg.inv`
> > 
> > Although the methods above only yield very slightly (around 1e-10) different S~−1 from the implemented approach. Manually setting the first step size for the RK solver also helps sometimes. Overall the problem is rather complicated.
> 
> That's weird. Will a larger epsilon solve this problem? Maybe we should try diffeq.jl/diffeqpy, because it supports more solvers (e.g. stiffness auto detections).

It turns out to be a bug - left out a `conj` when doing regularization. Sorry for all the fuss

liwt31:
Support GPU acceleration. The logic is the same as MPS/MPO. All TTS/TTO matrices are stored as NumPy array and they are converted to CuPy matrices when performing contraction.

liwt31:
Rename `TensorTreeState` and `TensorTreeOperator` to `TTNS` and `TTNO` for better consistency with existing literature

liwt31:
Add two-site tdvp-ps time evolution. Although I tried to implement the second-order symmetric Trotter splitting algorithm and the numerics are the same with the MPS implementation, the error is the same as first-order Trotter splitting algorithm. To test this, simply replace
```python
# in MPS language: left to right sweep
local_steps1 = _tdvp_ps2_recursion_forward(ttns.root, ttns, ttno, tte, coeff, tau / 2)
# in MPS language: right to left sweep
local_steps2 = _tdvp_ps2_recursion_backward(ttns.root, ttns, ttno, tte, coeff, tau / 2)
```
with
```python
# in MPS language: left to right sweep
local_steps1 = _tdvp_ps2_recursion_forward(ttns.root, ttns, ttno, tte, coeff, tau / 2)
# in MPS language: right to left sweep  // the following line is replaced as a second left-to-right sweep
local_steps2 = _tdvp_ps2_recursion_forward(ttns.root, ttns, ttno, tte, coeff, tau / 2)
```
Not sure if the observation is correct and the reason behind if so. Maybe this is a feature of 2-site TDVP-PS algorithm?

liwt31:
The branch is at a steady state. I'm planning to test the branch on several production-level tasks and maybe do a thorough refactoring to meet production requirements. Comments are still welcome but the PR will be closed.

PR #151, opened 2023-05-29
Revert dependency version to match python 3.7

atxy-blip:
Remaining warnings of `pytest`:

```shell
..\..\..\anaconda3\envs\reno\lib\site-packages\qutip\__init__.py:128
  ...\qutip\__init__.py:128: UserWarning: matplotlib not found: Graphics will not work.
    warnings.warn("matplotlib not found: Graphics will not work.")

renormalizer/mps/tests/test_mp.py::test_save_load
  ...\numpy\lib\npyio.py:719: VisibleDeprecationWarning: Creating an ndarray from ragged nested sequences (which is a list-or-tuple of lists-or-tuples-or ndarrays with different lengths or shapes) is deprecated. If you meant to do this, you must specify 'dtype=object' when creating the ndarray.
    val = np.asanyarray(val)
```

Tested on Windows 11 x86_64 with python=3.7, renormalizer=0.0.10.

### DIFF
--- a/renormalizer/transport/tests/test_kubo.py
+++ b/renormalizer/transport/tests/test_kubo.py
@@ -55,7 +55,7 @@ def get_qutip_holstein_kubo(model, temperature, time_series):
     j_oper = sum(terms).extract_states(qn_idx)
 
     # Add the negative sign because j is taken to be real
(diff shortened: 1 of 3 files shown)


liwt31:
Looks good to me, thanks!

PR #152, opened 2023-06-10
mpo dump remove warnings and future errors

jiangtong1000:
Starting from `numpy==1.24.0`, the occurrence of the `VisibleDeprecationWarning` when trying to save a list of lists with varying lengths has been rectified, resulting in a `ValueError` instead. So I made some changes here.


atxy-blip (commented):
The fix works for me.

liwt31:
LGTM,  thanks!

PR #154, opened 2023-08-05
Add functionality of MPO with block diagonal form

SUSYUSTC:
Currently if one add two MPOs by `mpo = mpo1 + mpo2` it creates zeros at off-diagonal blocks. In case the final MPO is composed of many small MPOs the computational efficiency is slow. I'll probably add another PR later for diagonal potential MPO, in which case the final MPO is composed of a dense kinetic block and a diagonal potential MPO block.

This pull request adds a `StackedMpo` class to improve the efficiency by calculating the Hamiltonians corresponding to each MPO part separately and then adding the Hamiltonians. It works for both the direct and iterative approach.

Example:
```
// original way: mpo = mpo_k + mpo_p
mpo = StackedMpo([mpo_k, mpo_p])
result = optimize_mps(mps.copy(), mpo)  // expect the same result
```


jjren:
Excellent PR and many thanks to Jiace.
It is better to add a test in `mps/tests/test_gs.py`. For example, in test_optimization, you can simply add the two mpos together and check if the energy is doubled.

SUSYUSTC:
Test already added.

jjren (approved):

atxy-blip:
The new function looks good to me. Many thanks for your work!

Meanwhile, personally speaking I prefer well-documented code, so maybe you could add new comments and make changes to the original ones accordingly. For example, the following line should contain the new class:

<https://github.com/shuaigroup/Renormalizer/blob/9706a29106f0b8213a60bfa18d9f1fa1bcec03f8/renormalizer/mps/gs.py#L55>

SUSYUSTC:
Comments & types added

PR #158, opened 2023-08-11
a trick but greatly accelerate h_qc construction.

jiangtong1000:
When constructing the qc mpo, simply discard useless loops which are very time consuming by picking only non-zero integrals.
Take the example of H-50 (100 spin orbitals), this trick speeds up thousands of times.

### DIFF
--- a/renormalizer/model/h_qc.py
+++ b/renormalizer/model/h_qc.py
@@ -143,15 +143,16 @@ def qc_model(h1e, h2e, conserve_qn=True):
         return Op.product(new_ops)
 
     # 1-e terms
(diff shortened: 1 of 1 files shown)


liwt31:
Cool! Can we use a threshold such as 1e-12 rather than exactly 0? Since they are float numbers

jiangtong1000:
> Cool! Can we use a threshold such as 1e-12 rather than exactly 0? Since they are float numbers

yea, but maybe it's better to let the screening of integrals happen outside the source code. (set by User)

liwt31:
Makes sense. LGTM, thanks!

liwt31 (approved):

PR #162, opened 2023-09-13
Enhance mps/mpo reproducibility; Print git hash for easier debugging;…

jiangtong1000:
- Enhance mps/mpo reproducibility; Currently, even with preset random seed, the MPO or the subsequent ground state MPS will be different for two runs. The MPO is different because the Op order can be different before doing Hopcroft-Corp; The MPS can be different because the sign carried by the eigenvectors.  This does not influence the final energy, but keeping everything reproducible could be important at sometime, especially for debugging purpose 
- The git hash information is printed every time running the package, for easier debugging
- Fix mp dump/load

### DIFF
--- a/renormalizer/mps/gs.py
+++ b/renormalizer/mps/gs.py
@@ -368,9 +368,11 @@ def sign_fix(c, nroots):
         if isinstance(c, list):
             return [ci / np.sign(ci[np.abs(ci).argmax()]) for ci in c]
         else:
-            return c / np.sign(np.max(np.abs(c), axis=0))
+            idx = np.abs(c).argmax(axis=0)
+            return c / np.sign(c[idx, range(c.shape[1])])
     else:
         return c / np.sign(c[np.abs(c).argmax()])
+
 
 def eigh_direct(
     mps: Mps,
--- a/renormalizer/mps/gs.py
+++ b/renormalizer/mps/gs.py
@@ -363,6 +363,15 @@ def get_ham_direct(
     return ham
 
 
+def sign_fix(c, nroots):
+    if nroots > 1:
+        if isinstance(c, list):
+            return [ci / np.sign(np.max(np.abs(ci))) for ci in c]
+        else:
+            return c / np.sign(np.max(c, axis=0))
+    else:
+        return c / np.sign(np.max(c))
+
 def eigh_direct(
     mps: Mps,
     qn_mask: np.ndarray,
@@ -383,11 +392,11 @@ def eigh_direct(
     nroots = mps.optimize_config.nroots
     if nroots == 1:
         e = w[0]
-        c = v[:, 0] / np.sign(np.max(v[:, 0]))
+        c = v[:, 0]
     else:
         e = w[:nroots]
-        c = [v[:, iroot] / np.sign(np.max(v[:, iroot])) for iroot in range(min(nroots, v.shape[1]))]
-    return e, c
+        c = [v[:, iroot] for iroot in range(min(nroots, v.shape[1]))]
+    return e, sign_fix(c, nroots)
 
 
 def get_ham_iterative(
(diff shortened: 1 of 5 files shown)


liwt31:
I really like this commit but there are some technical problems. Please see the comments

renormalizer/mps/symbolic_mpo.py
```diff
@@ -136,10 +136,11 @@ def construct_symbolic_mpo(table, factor, algo="Hopcroft-Karp"):
 
     # translate the symbolic operator table to an easy to manipulate numpy array
     table = np.array(table)
+
     # unique operators with DoF names taken into consideration
     # The inclusion of DoF names is necessary for multi-dof basis.
-    unique_op: Set[Op] = set(table.ravel())
-
+    unique_op = OrderedDict.fromkeys(table.ravel())
```
liwt31, line 142: Care to add a line of document here to explain why not use `set` directly?

renormalizer/mps/gs.py
```diff
@@ -383,10 +383,10 @@ def eigh_direct(
     nroots = mps.optimize_config.nroots
     if nroots == 1:
         e = w[0]
-        c = v[:, 0]
+        c = v[:, 0] / np.sign(np.max(v[:, 0]))
```
liwt31, line 386: I don't think this works as expected. You might first want to take absolute values before calling `max`. Also, this "canonicalize" logic is used below. Maybe it's better to wrap it as a utility function

renormalizer/mps/gs.py
```diff
@@ -556,4 +556,7 @@ def hop(x):
     else:
         assert False
     logger.debug(f"use {algo}, HC hops: {count}")
-    return e, c
+    if nroots == 1:
+        return e, c/np.sign(np.max(c))
+    else:
+        return e, c/np.sign(np.max(c, axis=0))
```
liwt31, line 562: This might cause the bug in CI. Taking max along the 0th axis will eliminate the 0th axis, resulting a vector with size N, where N is the dimension of c.

renormalizer/mps/gs.py
jiangtong1000: util function added and bug fixed

renormalizer/mps/gs.py
```diff
@@ -363,6 +363,15 @@ def get_ham_direct(
     return ham
 
 
+def sign_fix(c, nroots):
+    if nroots > 1:
+        if isinstance(c, list):
+            return [ci / np.sign(np.max(np.abs(ci))) for ci in c]
```
liwt31, line 369: 1. Taking abs loses the sign information
2. Add abs for the following branches as well?

jiangtong1000: 1. yea you are right. my mistake
2. what do you mean the following branches?

liwt31: when c is not a list and when nroots == 1

jiangtong1000: got it!

renormalizer/mps/gs.py
```diff
@@ -366,11 +366,11 @@ def get_ham_direct(
 def sign_fix(c, nroots):
     if nroots > 1:
         if isinstance(c, list):
-            return [ci / np.sign(np.max(np.abs(ci))) for ci in c]
+            return [ci / np.sign(ci[np.abs(ci).argmax()]) for ci in c]
         else:
-            return c / np.sign(np.max(c, axis=0))
+            return c / np.sign(np.max(np.abs(c), axis=0))
```
liwt31, line 371: It seems in this branch it still does not work properly?

jiangtong1000: thanks for pointing out! Now I think it is resolved.

PR #163, opened 2023-09-21
stacked qc mpo for larger scale qc calculation

jiangtong1000:
This PR is a follow up of #154 
Could improve the qc speed. The qc mpo is added by N smaller mpos with bond dim of O(N).

### DIFF
--- a/renormalizer/model/h_qc.py
+++ b/renormalizer/model/h_qc.py
@@ -67,7 +67,7 @@ def int_to_h(h, eri):
 
     return sh, aseri
 
-def qc_model(h1e, h2e, conserve_qn=True):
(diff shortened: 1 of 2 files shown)


liwt31:
Looks great! Is it reported in the literature? And how fast is it for typical cases?

liwt31 (approved):

jiangtong1000:
> Looks great! Is it reported in the literature? And how fast is it for typical cases?

yea I think this is a well known trick, see Eq 6 of (https://arxiv.org/abs/1703.04789)

jiangtong1000:
@liwt31 I did not test with typical systems yet, but in principle I think it can speed up N times? if we consider the step of updating the Environment

PR #169, opened 2024-05-22
Add DMRG pyscf solver interface

liwt31:
Allow renormalizer to interface with PySCF as a FCI solver. Ab initio calculation with renormalizer can now be performed using a few lines of code by PySCF.


jjren (approved):

jjren:
Good PR. It seems that the default tolerance of np.testing.assert_allclose is too strict.

liwt31:
> Good PR. It seems that the default tolerance of np.testing.assert_allclose is too strict.

Thanks for pointing out. It'll be fixed in the coming commits

PR #171, opened 2024-07-14
update: 2dof-entropy performance 

ansatzX:
past:
put dofs in calc-function one by one, build tree mutil-times

now
put all dofs in calc-function, build tree only once

add rdm input as entropy-calc function optional input, to imporve performance when we need it to be stored.


liwt31:
Nice improvement and thanks for all the doc/typo updates!

PR #172, opened 2024-08-24
optimize mpo construction

jjren:


liwt31:
Great work - could be useful in many cases

PR #173, opened 2024-11-15
add a tutorial example for SSH model ground state

jiangtong1000:


liwt31:
Thanks! Please update `example/run.sh` to make sure the file is tested during CI

liwt31:
Some of the tests failed but it's more like minor numerical errors. Thanks Jiang!

Issue #174, opened 2025-05-07
Solve deprecation warning on pip installs with a `pyproject.toml` file


atxy-blip:
During a very recent installation of the `renormalizer` package, I encountered the following deprecation message 

> DEPRECATION: Building 'renormalizer' using the legacy setup.py bdist_wheel mechanism, which will be removed in a future version. pip 25.3 will enforce this behaviour change. A possible replacement is to use the standardized build interface by setting the `--use-pep517` option, (possibly combined with `--no-build-isolation`), or adding a `pyproject.toml` file to the source tree of 'renormalizer'. Discussion can be found at https://github.com/pypa/pip/issues/6334

It seems that `pip` will no longer support installation from `setup.py` and `requirements.txt`. Instead, a standardized `pyproject.toml` file is preferred, which appears in other projects like [PySCF](https://github.com/pyscf/pyscf/blob/master/pyproject.toml). 

The following operation procedure is recommended for uploading the package to PyPI:

1. Ensure you have the latest versions of setuptools, wheel, and twine installed:
```sh
pip install --upgrade setuptools wheel twine
```
2. Build the distribution files (source and wheel):
```sh
python -m build
This will create the distribution files in the "dist/" directory.
```

3. Check the generated distribution files:
```sh
ls dist/
tar -xzf dist/renormalizer-0.0.11.tar.gz
cat renormalizer-0.0.11/PKG-INFO
```

4. Upload the package to PyPI using Twine:
```sh
twine upload dist/*
```

5. If you want to test the upload process, use TestPyPI:
```sh
twine upload --repository testpypi dist/*
```

6. Test the package installation from TestPyPI:
```sh
pip install --index-url https://test.pypi.org/simple/ renormalizer
```

I have managed to compose such a file, which will be added to the repo in a later PR. A test upload following above procedure can be reviewed at [here](https://test.pypi.org/project/renormalizer/0.0.11/).

liwt31:
@atxy-blip please close this issue if the PR solves the problem

PR #175, opened 2025-05-07
Add `pyproject.toml`

atxy-blip:
Corresponding to issue #174.

I am not certain whether to use the `README.md` in the repo as the PyPI homepage description file.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,0 +1,50 @@
+[project]
+name = "renormalizer"
+version = "0.0.11"
+# dynamic = ["version"]
+description = "Renormalizer is a Python based tensor network package with a special focus on electron-phonon quantum dynamics."
+
+authors = [
+    { name = "Weitang Li",     email = "liwt31@gmail.com"                 },
+    { name = "Jiajun Ren",     email = "jjren@bnu.edu.cn"                 },
+    { name = "Tong Jiang",     email = "tj.theochem@gmail.com"            },
+    { name = "Yuanheng Wang",  email = "yuanhwangtheo@gmail.com"          },
+    { name = "Yu Xiong",       email = "y-xiong22@mails.tsinghua.edu.cn"  },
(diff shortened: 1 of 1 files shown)


liwt31:
Thanks for the contribution. Overall LGTM, but I suggest adding more authors to the file. As we do not have a lot of contributors anyway, we may add all contributors shown on the repo index page.

atxy-blip:
> we may add all contributors shown on the repo index page

The `author/maintainer` field provided by [`PEP 621`](https://peps.python.org/pep-0621/#authors-maintainers) is a little bit ambigueous in its definition. From the suggestion of AI,
- *authors* are project founders or core developers with structural contributions
- *maintainers* are repository admins in charge of merge rights, release management, etc

PR contributors, without responsibilities for the project, are not applicable to both of these. Therefore I deem it proper to list @liwt31 alone in `pyproject.toml`.

On the other hand, one of the recommended approaches is to give them credit in `README.md`. I will push a separate commit for this later.

jiangtong1000:
The goal of this package is to be community-driven, and I believe it’s important not to impose too strict a definition here. My advice is to be more mindful of the social side of open-source. In case you didn’t notice, this package originated from the prototype `ephMPS`. Also, contributor recognition doesn’t necessarily need to be limited to the `README.md`(could be redundant), since GitHub already tracks contributors. If you truly want to highlight contributions, you could tie names directly to specific functions or code blocks, as was done in other package, such as https://github.com/block-hczhai/block2-preview

liwt31:
@atxy-blip I agree with @jiangtong1000. Please revise accordingly. AI suggestions are not strong arguments and can't replace original thinking. The PEP is actually quite clear: "The exact meaning (of the authors) is open to interpretation". We include the authors as we see fit.

atxy-blip:
I have added all possible contributors in a new commit, according to the suggestions above.

liwt31:
Great! Thanks for the contribution and looking forward to more!

